PR #4, opened 2019-01-30
Pep-8 corrections

GuillaumeAllain:
I propose some PEP-8 corrections, some minor and some major.

The major one is to switch from tabs to spaces. From PEP-8 recommendations, the use of spaces is preferred to keep the files more clean.

The rest is mostly adding spaces and compressing the lines into something less wide.


dccote:
We need two things to move forward:
1) is there a script that validates PEP8 syntax?
2) instructions for a few key editors (sublime text, emacs, etc?)

GuillaumeAllain:
Most modern python IDE support real-time syntax and error checking.
pycodestyle can be used with simpler text editing software (Atom, Sublime Text, probably emacs).

dccote:
OK, please write a small document in Markdown format to explain to any contributor how to validate their code for PEP8 compliance. I can then add it to README or elsewhere.

dccote:
In fact, probably simpler if I pep8 the current master....

dccote:
I modified the code with autopep8, so my master is now following pycodingstyle. Thanks.

PR #20, opened 2019-02-20
Francois lens and obj hi lo system

FrancoisCoteIng:
I created a Edmund.py file that contains 2 ACH lens: 100mm and 200mm focal distances, both 75mm in diameter. The info of each lens were taken on Edmund optics website, the url is in the python file. The effective focal length matches the one calculated for each lens (within 1% tolerance). Their callable name is in the format "AC{diameter}_{focalEFF}". Their stock number is in comments in each corresponding class.

I added the MVPlapo2XC objective from Olympus in the objectives.py file. An example path was also added in that file.


dccote:
Moved things around significantly to simplify the internal structure: objective.py and lens.py have been combined, and vendors have their own files which gets imported by default with its namespace.
Hence, to use a Olympus objective, you use: olympus.LUMPlanFL40X(), to use an Edmund Optics lens, you use: eo.PN_33_921() (PN is for part number because the class name cannot start with a digit).

dccote:
All examples from simple module files were removed and moved to __main__.py which is the real example file.

PR #44, opened 2020-05-05
Added possibility for multiprocessing independent of os

antho214:
Two methods were added to the `Matrix` class.

`Matrix. traceManyThroughInMultiprocess()` 
This method simply uses the `multiprocessing.Pool` function to map the calculation of ray trace across multiple processes. 

`Matrix. traceManyThroughInMultiprocessChunks()`
This method also uses the `multiprocessing.Pool` function, but can be faster in some cases. For example, when using `RandomRays` as input. In this function, we _manually_ split the rays in equal parts and send each part to its dedicated process.


raytracing/matrix.py
```diff
@@ -378,6 +380,62 @@ def _waitForFile(self, filePath, timeout=5):
             oldSize = currentSize
         time.sleep(0.1)
 
+    def traceManyThroughInMultiprocess(self, inputRays, processes=4):
+        """
+        Naive implementation of multiprocessing on `inputRays`.
+        """
+        with multiprocessing.Pool(processes=processes) as pool:
+            outputRays = pool.map(self.traceThrough, inputRays)
```
dccote, line 388: This here does not work because different processes access inputRays at once, and each access calls __iter__, which modifies `self.iteration`.  Sometimes, the clash, and the computation never stops.  This happens to me frequently.  This is because RandomRays() is not thread-safe: we must lock access to the __iter__ function OR expand the list explicitly.  

I looked at using Lock() but we can't do it simply because the lock cannot be used across processes (they use a file to exchange information, like I did originally but better). Therefore, I simply expanded the list explicitly to keep the code simple.

raytracing/matrix.py
```diff
@@ -378,6 +380,62 @@ def _waitForFile(self, filePath, timeout=5):
...
+        return Rays(rays=outputRays)
+
+    def traceManyThroughInMultiprocessChunks(self, inputRays, processes=4):
```
dccote, line 391: I find this code to difficult to read for non-experts even though I like the idea of using `traceManyTrough()` instead in `traceTrough()`. Also, if we create new dynamic classes, this will fail because e we check for a specific class.  This is not proper object-oriented programming because implementation details appear in the function.  It is better to avoid any specifics.

I found simple code to split my own array into smaller arrays easily at https://stackoverflow.com/questions/752308/split-list-into-smaller-lists-split-in-half

```
        theExplicitList = list(inputRays) 
        manyInputRays = [theList[i::processes] for i in range(processes)]
```

so I do that in the main function now, which I renamed again "inParallel".

dccote:
I don't see any benefit from chunk/no chunk, but I will leave the code there for now. I am merging it.

PR #56, opened 2020-05-06
Found a solution for matplotlib to allow ctrl-C in script window

dccote:
Finally found a solution to matplotlib ctrl-C disappearance.

It is fairly simple: use plt.draw() instead of plt.show(), loop and look for interrupts. Done!

Solutions was found here: https://stackoverflow.com/questions/28269157/plotting-in-a-non-blocking-way-with-matplotlib

### DIFF
--- a/raytracing/imagingpath.py
+++ b/raytracing/imagingpath.py
@@ -7,7 +7,7 @@ import matplotlib.pyplot as plt
 import matplotlib.patches as patches
 import matplotlib.path as mpath
 import matplotlib.transforms as transforms
-
+import sys
 
 class ImagingPath(MatrixGroup):
     """ImagingPath: the main class of the module, allowing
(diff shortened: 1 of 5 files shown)


dccote:
Ok, I think Ctrl-C on windows is not SIGINT but SIGBREAK.  So it appears that Ctrl-C is not KeyboardInterrupt, maybe we should catch "any" other Interrupt? Modify the code for except: with no specific interrupt, what do you get?

GabGabG:
It does the same thing: doesn't close the window (nor the program) and keeps popping new ones

dccote:
Ok, so let's put a check for Windows before the call.  If it is Windows, then we will use plt.show() and the problem will remain.  At least for Unix it will be solved.  We will create a new issue only for Windows... unless we find the solution tonight.

dccote:
New solution pushed to repository: check for windows and others. Use separate solutions.

antho214:
Running from the ctrlCMatplotlib branch, after opening a matplotlib figure when I click the terminal the focus shifts to the terminal, but the matplotlib figure stays on top. In fact it always stays on top whatever the other application that is selected. I can then hit Ctrl-C from the terminal and the figure disappears. However, it is an unexpected outcome.

record

dccote:
Yes I see that too but at least Ctrl c works...! Let’s merge.

PR #57, opened 2020-05-07
Ignore .idea folder

GabGabG:
.idea folder is a folder created by JetBrains IDEs (like PyCharm) and contains some "private" user specific information. Also, some files contained within the folder conflict with other users' files.

I added the whole folder in the .gitignore file. Github desktop (or just command line git) should not show files related to that folder anymore when ready to commit.

It is tricky dealing with ignoring files (or whole directories) when they are already commited (like .idea right now), so we have to be cautious to not commit them anymore, even if they are in the gitignore. I hope this will solve this annoying problem of conflicts and unimportant files commit.


dccote:
I understand the solution but it is too drastic: we still want to provide a good skeleton for the PyCharm project, it is in fact quite difficult to set up sometimes!  So we can use JetBrains.gitignore that I found at : https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore
this will ignore user files (breakpoints, etc...) but keep the project as a whole I think.  I am including it in gitignore now to see what happens.

By then deleting and reimporting the project, we can clearly keep only the few critical files:
image

PR #61, opened 2020-05-07
correction issue46

FrancoisCoteIng:
The property showElementLabels was not connected to any function. In the draw functions, showLabels was always set to True and the class property was not interrogated.


dccote:
Do you have quick test function that show the difference in both codes?

FrancoisCoteIng:
from raytracing import *
import raytracing.thorlabs as thorlabs
import raytracing.eo as eo
import raytracing.olympus as olympus

path=ImagingPath()
path.showElementLabels=False
path.append(Space(d=100,label="Space1"))
path.append(Lens(f=50,diameter=25.4,label="Lens"))
path.append(Space(d=100,label="Space2"))
path.display()

PR #65, opened 2020-05-12
Changed printed warnings to real warings

GabGabG:
Previous warnings were only printed with a normal print. I changed it for proper warnings (with red font and warning type). We are also able to test if warnings are really sent with unit tests.

### DIFF
--- a/.idea/RayTracing.iml
+++ b/.idea/RayTracing.iml
@@ -2,7 +2,7 @@
 <module type="PYTHON_MODULE" version="4">
   <component name="NewModuleRootManager">
     <content url="file://$MODULE_DIR$" />
-    <orderEntry type="jdk" jdkName="Python 3.7" jdkType="Python SDK" />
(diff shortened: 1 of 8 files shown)


dccote:
I get a weird `warnings.warn(msg, UserWarning)` line before the actual message.

dccote:
I just noticed that I don't get any colouring on my shell terminal (zsh)

GabGabG:
> ```
>   warnings.warn(msg, UserWarning)
> /Users/dccote/GitHub/RayTracing/raytracing/matrixgroup.py:37: UserWarning: Mismatch of indices between element 
> ```
@dccote :
The first line should now be removed, I had to change the format of warnings.

For the color, it works in the PyCharm console, but not in the comand prompt of Windows. Even exceptions don't show in color. It is quite tricky to have colors in terminals, but there are some modules able to do it. As usual, there are tricks that can be used but they don't work quite well on Windows.

Issue #48, opened 2020-05-06
Object and Image arrows often look weird when object/image very small


dccote:
We need an intelligent way of showing very small objects.

JLBegin:
- The size of the arrow heads (and aperture ticks) should scale with the display's FOV and the arrow's relative length.
- A constant blank space (relative to the FOV) should also be left in the Y-axis of the display.
- The default fan angle could also scale with the display.
- Are the object labels too small ?
- When an object is too small to be noticed, we could add an additional label or something else (need more thinking).
- Do we want all the traced rays to be present inside the display (y-scale)? or just the objects?

Issue #16, opened 2019-02-15
The y-scale is not always set appropriately when the elements have infinite diameters


dccote:
The y-scale is not always set appropriately when the elements have infinite diameters: the rays will go beyond the element drawn on the figure. It should be simple to figure out a way to set the displayHalfHeight appropriately based on the highest ray.

dccote:
This is overlapping with issue #76 and could be fixed at the same time.

JLBegin:
@dccote I think this issue should have been closed with #76 , no ?

PR #76, opened 2020-05-13
Improved display

JLBegin:
Even though I don't think this pull request is ready to be merged, I thought I would open it to discuss the changes and what is left to be done. 

## Summary
The commits tell the story, but here's a **summary of the changes**:
- The `displayRange` required for `ImagingPath.createRayTracePlot` is now a property (and tested).
- The `displayRange` (y-axis limits) is defined as 1.5 times the largest _object_ drawn (this includes the elements, the object and the images) which creates a 50% blank space in the y-axis in total (+/- 25%). 
- In order to consider the image height for the `displayRange` (see example 2), ~~I added the class variable `ImagingPath.images` containing a list of (imagePosition, magnification) couples (I know this might not be the ideal solution), and~~ I split `drawImages()` into `intermediateConjugates()` in Matrix and `drawImages()` in ImagingPath so that .createRayTracePlot first takes into account the image heights to properly set the `displayRange`. 
- Replaced the `Matrix.axesToDataScaling` (which would sometimes fail) with a new method called `Matrix.axesToDataScale` [see their relative _temporary_ test files]
- The arrows' size and the aperture stops now correctly scale with the relative size of their _object_ (lens, object, images) to the display (see issue #48)

## Edits
- The arrows' width is now unnafected by matplotlib's zoom. In order to do that, I moved the last calls of `createRayTracePlot` to a `drawDisplayObjects` function that creates the matplotlib artists and patches of all elements except rays. The `drawDisplayObjects` is then also called inside a callback function `updateDisplay` which first clears all the axes's drawn objects.

## Questions
- I arbitrarily set a default size and a scaling factor for the arrows. Is it too big or too small? Does it scales well?
- Is the y-axis blank space alright?
- Do we want to include the rays inside the displayRange or we allow them to go out? (see example 1)
- Do we want to add special annotations for objects too small to be noticed at first ? (see example 3)
- Should we also scale the fontsize of labels and annotations?
- Should we also scale the default fan angle of the traced rays ? (see how the default value affects example 1 vs example 2)

## TODO
- > We should always scale the Lenses and other objects (if they have infinite diameters) to be as large (and a bit more) than the farthest rays. This way, they wouldn't look weird like Example 1, and everything would fall into place naturally.
**Edit**: now implemented for lens (only)
- > I am hoping we will stop using objectHeight and fanAngle to replace them with principal rays when they exist, unless the user explicitly sets an objectHeight and fanAngle. Also, we will need to move towards the use of Rays() instead: they would allow all the flexibility we need.

### Example 1
image

### Example 2
image

### Example 3a
image

### Example 3b (close-up)
image


raytracing/imagingpath.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, elements=None, label=""):
         self.showPointsOfInterest = True
         self.showPointsOfInterestLabels = True
         self.showPlanesAcrossPointsOfInterest = True
+        self.images = None
```
dccote, line 42: Recurring comment here: I think to keep the nomenclature standard, we should refer to these as something like `intermediateForwardConjugates` or `intermediateForwardConjugatePlanes` or maybe just `intermediateConjugates`. because the class `Matrix` already has `forwardConjugate` and `backwardConjugate` to find the last conjugate plane after or the first before the element.  `Matrix` has now internal structure, so it makes no sense to have `self.intermediateForwardConjugates` functions there.  However, `MatrixGroup` is the most general class after `Matrix` that has an internal structure , so we should probably move that function there (later?). A `MatrixGroup` does have intermediate conjugate planes, whether it is an imaging path or not.
Also, these planes exist regardless of what the object is (present, absent, etc...), so we should refer to them as `planes` more than `images`.

That being said, we should remove the intermediate variable that only adds complexity and chances of error (see below at `computeImages()`)

raytracing/imagingpath.py
```diff
@@ -299,6 +300,17 @@ def lagrangeInvariant(self, ray1=None, ray2=None, z=0):
 
         return super(ImagingPath, self).lagrangeInvariant(z=z, ray1=ray1, ray2=ray2)
 
+    @property
+    def displayRange(self):
```
dccote, line 304: We need a docstring even if the format is not finalized yet explaining the rationale for these choices.

raytracing/imagingpath.py
```diff
@@ -423,50 +433,62 @@ def save(self, filepath,
 
     def drawObject(self, axes):  # pragma: no cover
         """ Draw the object as defined by objectPosition, objectHeight """
-        (xScaling, yScaling) = self.axesToDataScaling(axes)
```
dccote, line 426: Not sure why `self.axesToDataScaling` is gone because it is not clear `axesScale()`refers to scaling with respect to what?  Since the drawing functions are in data space (i.e. they use the coordinates of the plot), then it sounds natural to keep it in the name.
Scaling in graphs can become fairly complicated very quickly with the many reference frames (window, figure, axes, bounds, data, normalized, etc...), I would prefer if we kept the source-destination frames of reference explicit in the functions.

raytracing/imagingpath.py
```diff
@@ -423,50 +433,62 @@ def save(self, filepath,
...
 
-    def drawImages(self, axes):  # pragma: no cover
-        """ Draw all images (real and virtual) of the object defined by 
-        objectPosition, objectHeight """
-
-        (xScaling, yScaling) = self.axesToDataScaling(axes)
-        arrowWidth = xScaling * 0.01
-        arrowHeight = yScaling * 0.03
-
+    def computeImages(self):
```
dccote, line 454: Let's rename this `intermediateForwardConjugates(self)` and return the array instead of setting the internal variable. Let's also drop the internal variable (`self.images`).  We'll just compute it every time because setting an internal variable on a computation "in the background" is not the best practice: what if someone tries to use images without calling `computeImages()`?  Since the computation is not expensive at all (not much is intensive in this module), let's recompute it every time.

raytracing/imagingpath.py
```diff
@@ -423,50 +433,62 @@ def save(self, filepath,
...
+                    images.append([imagePosition, magnification])
+        self.images = images
+
+    def drawImages(self, axes):  # pragma: no cover
+        """ Draw all images (real and virtual) of the object defined by 
+        objectPosition, objectHeight """
+
+        (xScaling, yScaling) = self.axesScale(axes)
+
+        for (imagePosition, magnification) in self.images:
```
dccote, line 474: Here, we should simply call `images = self.intermediateForwardConjugates()`

raytracing/matrix.py
```diff
@@ -746,7 +751,14 @@ def displayHalfHeight(self):
             halfHeight = self.apertureDiameter / 2.0  # real half height
         return halfHeight
 
+    def axesScale(self, axes):
+        xlims, ylims = axes.get_xbound(), axes.get_ybound()
+        xScale, yScale = xlims[1]-xlims[0], ylims[1]-ylims[0]
+        return xScale, yScale
+
     def axesToDataScaling(self, axes):
+        raise DeprecationWarning("Use .axesScale() instead")
```
dccote, line 760: See comment above regarding keeping the reference frames explicit in the function names.

raytracing/tests/testMatrixAxesToDataScaling.py
```diff
@@ -0,0 +1,76 @@
+import unittest
+from raytracing import *
+
+
+class TestMatrixAxesToDataScaling(unittest.TestCase):
+    def testWithEmptyAxes(self):
```
dccote, line 7: All these tests fail because of the deprecation warning.

dccote:
Another issue (for which I don't have an answer) is that if we zoom using the magnifying glass in the Matplotlib window, then the arrows look like crap. Thinking out loud, I think the arrows should be draw in "window" coordinates, that is, they should always be the same thickness and size regardless of the zoom level in Matplotlib.

dccote:
> * Do we want to include the rays inside the displayRange or we allow them to go out? (see example 1)

I think we should and at the same time but only indirectly? We should always scale the Lenses and other objects to be as large (and a bit more) than the farthest rays. This way, they wouldn't look weird like Example 1.

JLBegin:
> Another issue (for which I don't have an answer) is that if we zoom using the magnifying glass in the Matplotlib window, then the arrows look like crap. Thinking out loud, I think the arrows should be draw in "window" coordinates, that is, they should always be the same thickness and size regardless of the zoom level in Matplotlib.

Right. Like how line or markers work, where the zoom doesnt affect the drawing.

dccote:
> * Should we also scale the default fan angle of the traced rays ? (see how the default value affects example 1 vs example 2)

I am hoping we will stop using objectHeight and fanAngle to replace them with principal rays when they exist, unless the user explicitly sets an objectHeight and fanAngle.  Also, we will need to move towards the use of `Rays()` instead: they would allow all the flexibility we need.

JLBegin:
> > Another issue (for which I don't have an answer) is that if we zoom using the magnifying glass in the Matplotlib window, then the arrows look like crap. Thinking out loud, I think the arrows should be draw in "window" coordinates, that is, they should always be the same thickness and size regardless of the zoom level in Matplotlib.
> 
> Right. Like how line or markers work, where the zoom doesnt affect the drawing.

But if we draw a really small arrow, we won't be able to effectively zoom on it...

raytracing/imagingpath.py
JLBegin: We can keep the name axesToDataScaling. But to be clear, the function simply returns the dimensions of the display in data units. Also, since it's an axes property, i'm not sure that it should be defined in Matrix.

JLBegin:
I looked into getting the objects' size to not change when zooming. I know transforms are normally used, but it gets pretty complicated and not really doable if you want the object to not zoom in x-axis, but still pan around and rescale in y-axis. I think the easy way instead is to take the zoom callback to update the object's width (keeping auto position panning and auto y-axis scaling) since that's really our only problem.

JLBegin:
Small update to the drawing of infinite lens only. I set a minimum size with the highest ray if it's higher than 4 (could be removed but it's a more general matrix property). For now this feature is only for Lens. Linked issue #16 .
image

JLBegin:
Now considering ray traces in displayRange if it has infinite aperture. This lead to this kind of display for example 1
image

raytracing/imagingpath.py
dccote: So then the name is fine: we go from normalized axes coordinates to data coordinates. I don't remember all the details, but I believe there is a reference frame in normalized axes coordinates.

setup.py
```diff
@@ -42,9 +42,9 @@
     keywords='optics optical lens ray tracing matrix matrices aperture field stop\
     monte carlo design raytracing',
     packages=setuptools.find_packages(),
-    install_requires=['matplotlib'],
-    python_requires='>=3.6',
```
dccote, line 46: Are we suppose to say 3.6 here or 3? SHouldn't that be 3.6 ? where else would we say it?

raytracing/imagingpath.py
```diff
@@ -310,6 +310,13 @@ def displayRange(self):
             for (planePosition, magnification) in conjugates:
                 if displayRange < self.objectHeight * magnification:
                     displayRange = self.objectHeight * magnification
+
+        if axes is not None:
+            if len(axes.lines) != 0:
```
dccote, line 315: Ouch: a bit of a hack. We sometimes use lines for other things, this is not particularly safe or robust.

raytracing/imagingpath.py
```diff
@@ -310,6 +310,13 @@ def displayRange(self):
...
+                for rayTrace in axes.lines:
+                    if max(abs(rayTrace._y)) * 2 > displayRange:
+                        displayRange = max(abs(rayTrace._y)) * 2
+
         return displayRange
 
     def createFigure(self, limitObjectToFieldOfView=False,
```
dccote, line 322: Change of API. Tut, tut, tut... Here is a video of the last person who did that: https://youtu.be/GTQAXX08A-s?list=WL&t=21
We should discuss any change of API.

raytracing/imagingpath.py
dccote: Indicate FIXME: in a comment.

setup.py
JLBegin: This has to be 3.6 , a recent master merge changed it.

raytracing/imagingpath.py
PyMarc2: it's going down

JLBegin:
Alright I think it's ready to be merged as discussed @dccote. This pull request currently has the branch article-tutorial as destination.

JLBegin:
Alright I think it's ready to be merged as discussed @dccote. This pull request currently has the branch article-tutorial as destination. I will soon start a new branch for draw commands.

PR #83, opened 2020-05-13
setup missing packages + requirements.txt

PyMarc2:
for pypi users setup.py completion and requierement.txt for devloppers


.gitignore
```diff
@@ -187,3 +187,7 @@ fabric.properties
 
 # Android studio 3.1+ serialized cache file
 .idea/caches/build_file_checksums.ser
+
+.idea/RayTracing.iml
+
+.idea/misc.xml
```
JLBegin, line 193: +1 from me. 
And I would also add other.iml

image

PyMarc2: yeah, but is there any .idea file that we absolutely want to keep... ? I mean, they regenerate usually, i'm not a fan of synching the .idea as it alwaaayyyss changes and people can have their own settings

requirements.txt
```diff
@@ -0,0 +1,8 @@
+cycler==0.10.0
+env==0.1.0
```
dccote, line 2: I created a small env.py file, I don't think this is a real package....

requirements.txt
```diff
@@ -0,0 +1,8 @@
...
+kiwisolver==1.2.0
```
dccote, line 3: Where is this used?

requirements.txt
```diff
@@ -0,0 +1,8 @@
...
+matplotlib==3.2.1
+numpy==1.18.4
+pyparsing==2.4.7
```
dccote, line 6: pyparsing? Where is this used?

requirements.txt
```diff
@@ -0,0 +1,8 @@
...
+python-dateutil==2.8.1
+six==1.14.0
```
dccote, line 8: six? Where is this used?

requirements.txt
PyMarc2: Thse are basics packages installed with every virtual environnement

PyMarc2: We can remove them indeed. But they will reinstall automatically when using the venv std python library. I can dig deeper. I thought it was not worth it, but I will.

requirements.txt
PyMarc2: Six is a basic venv depency package. It is used to smooth out the differencies of a python 2.X interpereter vs a python 3.X interpreter

requirements.txt
PyMarc2: - These are all matplotlib dependencies

requirements.txt
PyMarc2: This is a matplotlib dependencie

requirements.txt
PyMarc2: Yeah I saw that, but I wasn't 100% sure because the env package exists and the env.py is located in test, which would need from .test import env to import, no ? I'l be looking into it.

.gitignore
PyMarc2: @dccote

dccote: It has been removed properly in one branch but I don't remember which one.

requirements.txt
dccote: Ok, I will clearly need to change its name.

dccote: Done.

setup.py
```diff
@@ -43,7 +43,7 @@
     keywords='optics optical lens ray tracing matrix matrices aperture field stop\
     monte carlo design raytracing',
     packages=setuptools.find_packages(),
-    install_requires=['matplotlib'],
+    install_requires=['matplotlib', 'numpy', 'env'],
```
dccote, line 46: remove env?

.gitignore
dccote: We are only keeping shared files in the .idea folder, none of the user files.

dccote: Take a look at the .gitignore.  There is a section at the bottom.  I am not sure what we should be doing with .iml files.

requirements.txt
```diff
@@ -0,0 +1,8 @@
+cycler==0.10.0
```
dccote, line 1: I am reading everywhere that requirements.txt are not needed?
https://medium.com/@boscacci/why-and-how-to-make-a-requirements-txt-f329c685181e
When is this file used really? If it is only for developers, we need to hide it better somewhere in the directories.

PyMarc2: requirements.txt are for developpers. not for the actual package, as i said in the commit comment

PyMarc2: I don,t really see anywhere to place it. It has always been in the main folder everytime I have seen projects with people working with venv. It seems that making a folder specifically for this file might be too much.

.gitignore
PyMarc2: Yeah on the .idea issue. do we make a separate branch to clean it properly ?

requirements.txt
JLBegin: or you build your dev environment with pip install raytracing (installs dependencies) then pip uninstall raytracing

PyMarc2: yeah, I think this is not really a dev environnement issue since it has only matplotlib and numpy,  but i don't understand the angst against venv or the need to hide requirements.txt, since it is common practice, which is the proper method i believe.

PyMarc2: Nonehteless, I'll remove the file.

PR #89, opened 2020-05-14
Added transferMatrix in ThickLens and DielectricSlab

GabGabG:
ThickLens and DielectricSlab now override transferMatrix. They return a new ThickLens or a new DielectricSlab if the distance is smaller than the physical length. Otherwise they return the current instance (self).


raytracing/matrix.py
```diff
@@ -1071,6 +1071,14 @@ def pointsOfInterest(self, z):
 
         return pointsOfInterest
 
+    def transferMatrix(self, upTo=float('+Inf')):
+        """ Returns a ThickLens() corresponding to a partial propagation
+        if the requested distance is smaller than the length of this element"""
+        if self.L <= upTo:
+            return self
+        else:
+            return ThickLens(self.n, self.R1, self.R2, upTo, self.apertureDiameter, self.label)
```
dccote, line 1080: You need to return a Matrix (or MatrixGroup) that has DielectricInterface + Space(appropriate length).

raytracing/matrix.py
```diff
@@ -1095,6 +1103,14 @@ def drawAt(self, z, axes, showLabels=False):  # pragma: no cover
                               clip_on=True)
         axes.add_patch(p)
 
+    def transferMatrix(self, upTo=float('+Inf')):
```
dccote, line 1106: You need to return a Matrix (or MatrixGroup) that has DielectricInterface (infinite radius) + Space(appropriate length).

Issue #79, opened 2020-05-13
The __main__ file is not particularly easy to read and is hard to maintain


dccote:
It would be nice and useful to have a function, in some sort of `PlotManager`, that would take code as an argument, run it, and print the code on screen with syntax highlighting. We could even have have the option of providing the code as a string or as a file on disk. This way, for instance, we could store all examples in the examples directory and access them directly.

PyMarc2:
Could we just add arguments to the argparse for the file and code management?
`python -m raytracing -f testRaytracing.py`
`python -m raytracing -s "ALL THE CODE AS A STRING with \n and stuff."`

and for the `python -m raytracing -e`, I agree that every example should be in the examples folder and a sort of `ExampleManager` would select the example to display from the file. Maybe even this object could manage the input from the argparse when we use -f or -s for example @dccote

PyMarc2:
However, at this point, if they have a code `test.py` that look like this:
`
path = ImagingPath()
    path.label = "Demo #1: lens f = 5cm, infinite diameter"
    path.append(Space(d=10))
    path.append(Lens(f=5))
    path.append(Space(d=10))
    path.display()
`
At this point, isn't easier AND clearer to just IMPORT the raytracing library components into their code `
test.py` ?

PyMarc2:
I think that segmentation of the example codes is useful: less messy and more easily understandable if you want to make a new example, you just add it in the example folder

PyMarc2:
ok. Here's the plan
Each example.py file is separated. The example comment is in docstrign instead.
There will be a ExampleManager which will generate the figure and save it in pdf.

Then, the code of that file will the converted to coloured HTML using the pygment module.
From the HTML output we will generate a PDF. Then we will juxtapose the 2 PDFs one over the other in a new PDF which will be displayed. Thus, the code will be highlighted and selectable.

PyMarc2:
See branch #issue79-mainEnhancement

PyMarc2:
![image](https://user-images.githubusercontent.com/31829033/81970689-eb8e9800-95ed-11ea-8854-d9bb5ada0c21.png)
This is the HTML file output for the code highlighting and selectability

PyMarc2:
The output seems leggit. Here's a quick view. Does it asthetically please you @dccote ?
![image](https://user-images.githubusercontent.com/31829033/82005033-5a8de000-9632-11ea-832b-10acb9e2c68d.png)

dccote:
[...]

PR #93, opened 2020-05-15
issue #79 aesthetics of example and __main__ cleanup + more maintanable

PyMarc2:
-The output are in HTML
-tried to be aesthetically pleasing codewise and examplewise
-the examples can be written -as such- in exampleX.py and it will generate it automatically.


PyMarc2:
Verify conflicts with .gitignore before merging...

raytracing/__main__.py
```diff
@@ -1,430 +1,123 @@
...
+        fig, axes = plt.subplots(figsize=(7, 5))
+        self.figureObject.createRayTracePlot(axes=axes)
+        with tempfile.TemporaryFile(suffix=".png") as tmpfile:
+            plt.savefig(tmpfile, format="png")
+            tmpfile.seek(0)
+            self.base64Figure = base64.b64encode(tmpfile.read()).decode()
+        self.HTMLFigure = ''''<div style="display:inline-block; vertical-align:top"><img class="icon" src="data:image/png;base64,{} "></div>'''.format(
+            self.base64Figure)
+
+    def generateHTMLHighlightedCode(self, code=""):
```
JLBegin, line 81: Do you plan on using this `code` argument ?

PyMarc2: It was for future proofing. Daniel talked about inline string code input. Even though i'm not sure about that, I kept it in mind that the argument of -s might be a string to highlight

raytracing/__main__.py
```diff
@@ -1,430 +1,123 @@
...
+        code = self.getExampleCode()
+        docString = self.getDocstring(code)
+        self.HTMLDescription = '''<div class="alert alert-success" role="alert"><h4 class="alert-heading">Example Description</h4><div style="word-wrap: break-word;
+   width: 600px;">{}</div></div>'''.format(docString)
+
+    def runExampleCode(self):
+        with patch('matplotlib.pyplot.show') as p:
+            global path
+            exec(open(self.selectedFile).read(), globals())
+            self.figureObject = path
```
JLBegin, line 99: I don't think we should call this ImagingPath a figureObject since it's not.

PyMarc2: very true.

raytracing/__main__.py
JLBegin: Ok. I guess this could enable a user to create it's own html examples...  in the future

PyMarc2:
Discovered error. in __main__ we arg is 'all', must make the index = i +1

raytracing/__main__.py
```diff
@@ -1,430 +1,123 @@
-from .imagingpath import *
-from .laserpath import *
+import argparse
+from pygments import highlight
+import pygments.lexers
+import pygments.formatters
+import webbrowser
+from unittest.mock import patch
+from matplotlib import pyplot as plt
+from pylatexenc import latexencode
```
JLBegin, line 8: requires package `pylatexenc`

PyMarc2: it does not, my bad

raytracing/__main__.py
```diff
@@ -1,430 +1,123 @@
...
-from .axicon import *
-import raytracing.thorlabs as thorlabs
-import raytracing.eo as eo
-import raytracing.olympus as olympus
 
-import argparse
 ap = argparse.ArgumentParser(prog='python -m raytracing')
-ap.add_argument("-e", "--examples", required=False, default='all', help="Specific example numbers, separated by a comma")
-
+ap.add_argument("-e", "--examples", required=False, default='', help="Specific example numbers, separated by a comma")
```
JLBegin, line 15: We could specify the allowed range, and catch a valueError if out of range

PyMarc2:
- The only non native new package is pygments, which is necessary anyway if the code has to be highlighted.
- It should now run because the pylatex module wasn't supposed to be there
- you believe it is less readable? sad, I think it is much better than 30 example in one document!
- yes the OS dependency is a problem to verify.
- yes I thought you would want the output to be in native matplotlib display
- I thought the examples gave exactly the same result as before, not sure I understand ?

dccote:
> * you believe it is less readable? sad, I think it is much better than 30 example in one document!

The `__main__` is less readable. Clean it up big time to quickly get the code from all example files in a simple loop on example dir.  We will also call the files with useful names.  It could be example-kohlerIllumination.py for instance.
We may want to move the folder inside raytracing, and provide the link to the user so they can open the directory.

I think it's important that the example file be standalone.  `__main__` will simply loop through them.

dccote:
We'll revisit this later, for now, let's close the Pull Request and keep the branch.

PR #99, opened 2020-05-15
try the change relative paths to RAW html links

PyMarc2:
Seems to be the consensus online as the main fix to the relative path or blob link problems


dccote:
I believe this was fixed, but only in tutorial branch.... Sorry.

dccote:
Was already solved but not noted as such.  My bad.

dccote:
it is not a solved issue.

dccote:
@pymarc2 Where do we stand with thIs? Do we need REadme.rst? How is it maintained?

dccote:
@PyMarc2 Where do we stand on this?

PyMarc2:
The .rst can be made automatic, but it isn't for now. but from what I'm reading, it isn't the problem. I've seen packages with .md with their images working. e.g tensorflow, tensorflow scientific. I'm really not sure yet, there's no pattern

PyMarc2:
Resolved (the problem works with the RAW links, it's just Pypi that takes time to update)

PR #100, opened 2020-05-17
New documentation platform

JLBegin:
Available at https://raytracing.readthedocs.io/

- Using [Sphinx](https://www.sphinx-doc.org/en/master/) documentation generator. This is easily maintained with 
  - The configuration file `conf.py` (set theme, extensions, etc). 
  - The `index.rst` referring to the different sections of the docs (2 for now) : `reference.rst` (API autodocs) and `README.md`. 
  - The API docs are generated from `reference.rst` which calls a custom template for auto-class documentation `_templates\autoClass.rst`. 
- Hosted on [ReadTheDocs](https://readthedocs.org/) which also automates the building.

The API documentation is automatically generated from the docstrings in the code. The docstrings have to follow the required [Google Docstring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) format.

More info in the [README](https://github.com/DCC-Lab/RayTracing/blob/documentation/docs/README.md)


ElaheParham:
It looks good. Thank you!
we need to talk about the details.
I suggest we include the following sections: installation (getting started), support (contribute),
 FAQ, about, examples

And the NumPy format seems standard for the details of the functions.

JLBegin:
> It looks good. Thank you!
> we need to talk about the details.
> I suggest we include the following sections: installation (getting started), support (contribute),
>  FAQ, about, examples
> 
> And the NumPy format seems standard for the details of the functions.

Yes, these sections are a must, but I thought I would keep the writing for another branch. 

For the docstrings, it supports both Google and Numpy formatting, but for clarity we should choose one of them. Google uses indentations to mark sections while Numpy uses underlines (with equal signs If I remember correctly). I think PEP recommends Google.

ElaheParham:
> > It looks good. Thank you!
> > we need to talk about the details.
> > I suggest we include the following sections: installation (getting started), support (contribute),
> > FAQ, about, examples
> > And the NumPy format seems standard for the details of the functions.
> 
> Yes, these sections are a must, but I thought I would keep the writing for another branch.
> 
> For the docstrings, it supports both Google and Numpy formatting, but for clarity we should choose one of them. Google uses indentations to mark sections while Numpy uses dashed lines.

we can look at the following two examples (I found them on the web) and see the differences:

**_example of google:_**
```
"""
This is an example of Google style.

Args:
    param1: This is the first param.
    param2: This is a second param.
Returns:
    This is a description of what is returned.

Raises:
    KeyError: Raises an exception.
"""
```


_**example of numpy:**_
```
"""
My numpydoc description of a kind
of very exhautive numpydoc format docstring.

Parameters
----------
first : array_like
    the 1st param name `first`
second :
    the 2nd param
third : {'value', 'other'}, optional
    the 3rd param, by default 'value'

Returns
-------
string
    a value in a string

Raises
------
KeyError
    when a key error
OtherError
    when an other error
"""
```
--------------------------------
between these two, personally I prefer Numpy because you can easily obtain more information by a glance at the document and I guess it is widely used by the scientific community.

I didn't see any recommendations from PEP. [This link](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#google-vs-numpy) is a comparison of them.
As it said :
> NumPy style tends to require more vertical space, whereas Google style tends to use more horizontal space. Google style tends to be easier to read for short and simple docstrings, whereas NumPy style tends be easier to read for long and in-depth docstrings.

I think if we want to include more details and information NumPy will be better and if we want to write a short description google style is the right choice.

JLBegin:
Indeed, the recommandation was from Khan Academy, not PEP. Numpy is clean, lets go with that.

Since we removed the base branch I will open a new pull request rebased on master. Daniel is putting us in charge of the docs so we should merge this soon and start another branch for the writing.

PR #114, opened 2020-05-19
Matrix and its subclasses are fully tested.

GabGabG:
Display methods are not covered and <del>`axesToDataScale()` is not tested because @JLBegin was working on it</del> `axesToDataScale()` is tested because @JLBegin 's new version is merged in master. There are still deactivated tests.


JLBegin:
> Display methods are not covered and ~`axesToDataScale()` is not tested because @JLBegin was working on it~ `axesToDataScale()` is tested because @JLBegin 's new version is merged in master. There are still deactivated tests.

Yes and most display methods and `axesToDataScale` will soon be moved to a LayoutManager class anyway.

PR #116, opened 2020-05-19
Trace many through now accepts iterables (tuple, arrays, etc)

GabGabG:
It was only accepting lists and Rays objects, but arrays (numpy) and tuples are also decent types that should be accepted. It was throwing an exception which was not very helpful to know.


GabGabG:
Oops: `str` is an iterable, so it is not done yet! (we don't want `str`...)

GabGabG:
@dccote I think checking if it is iterable is good for `traceManyThrough`, but further research (if it is an iterable or `Ray` object) should be done in `Rays`?

GabGabG:
For now, `traceManyThrough()` checks if the input is iterable.

PR #123, opened 2020-05-20
Update __main__.py

dccote:
Relative imports have been added for `eo`, `thorlabs` and `olympus` lenses instead of absolute imports.  The old syntax was implicitly relative in Python 2 only.


dccote:
Duplicate

Issue #90, opened 2020-05-14
ImagingPath and LaserPath methods in MatrixGroup raise exception.


GabGabG:
They respectively return a ImagingPath instance and a LaserPath instance, but those classes are in other files not imported. It is also worth mentioning that they both inherit from MatrixGroup, so it is kind of incoherent to use them in MatrixGroup.

Question:
Are those methods really necessary in MatrixGroup?
- Yes: We have to find a way to import them without having a circular import problem.
- No: We should remove them.

dccote:
HUmmm..... Hard one: they are convenient if you work with MatrixGroup and then decide to look at at as an imagingPath or LaserPath().  But you are right, it does not belong in the parent class. 
We should instead have a conversation niece init in `ImagingPath` and `LaserPath` that can take a `MatrixGroup`

PR #126, opened 2020-05-20
Tests matrix group 98% coverage

GabGabG:
There are two functions not covered: `ImagingPath()` and `LaserPath()`. The fact that they are not tested is related to issue #90 . There is also one missing condition branch in `intermediateConjugates()`. It cannot be tested because the condition is `if distance is not None` and `distance` cannot be `None`. It can be infinity, **perhaps we should check if the conjugate is at infinity?**


dccote:
I am wondering if we should return a conjugateMatrix with +inf elements? Technically, it does exist as a limit. I wonder how well the rest of functions would behave?

dccote:
I added tests for limits to see if we could keep +inf and -inf as values.  I am still not sure, but we can certainly handle well defined limits (inf * inf for instance), this is one automatically. I think the problem we will have is with inf * 0, which will happen.  I am looking into that.

GabGabG:
> I think the problem we will have is with inf * 0, which will happen. I am looking into that.
Yes, in python, inf * 0 is nan, this is kind of weird that this indetermination is considered, but 1^inf and inf^inf is not.

dccote:
> > I think the problem we will have is with inf * 0, which will happen. I am looking into that.
> 
> Yes, in python, inf * 0 is nan, this is kind of weird that this indetermination is considered, but 1^inf and inf^inf is not.


Interesting for 1^inf, but not surprising: 1.0 is exact in binary, so it simply handles it directly I guess with (1)^anything always returns 1.  Wouldn't be so nice if 1.0 was in fact 1.00000001 or 0.9999999.  As a result of calculation, this would go one way or the other.

PR #130, opened 2020-05-20
Removed useless redefinition of len() and 'if self.rays is None, if self.rays is not None' verifications

### LINKED ISSUE #129: Rays has two len methods
The class `Rays` has two `__len__()`. They mostly share the same code, but one check if `self.rays is None`, which is useless, because `rays` can't be `None`: it is set as `[]` in the constructor if no initial rays are given.

GabGabG:
We don't need to make sure `self.rays is not None` since it can't be. 
Fixes #129.

### DIFF
--- a/raytracing/rays.py
+++ b/raytracing/rays.py
@@ -133,9 +133,6 @@ class Rays:
         return self
 
     def __next__(self) -> Ray:
-        if self.rays is None:
-            raise StopIteration
-
         if self.iteration < len(self.rays):
             ray = self.rays[self.iteration]
             self.iteration += 1
@@ -147,8 +144,7 @@ class Rays:
         return self.rays[item]
 
     def append(self, ray):
-        if self.rays is not None:
-            self.rays.append(ray)
+        self.rays.append(ray)
 
         # Invalidate cached values
         self._yValues = None
--- a/raytracing/rays.py
+++ b/raytracing/rays.py
@@ -11,13 +11,14 @@ or output plane) for ImagingPath, MatrixGroup or any tracing function.
 Subclasses can provide a computed ray for Monte Carlo simulation.
 """
 
+
 class Rays:
     def __init__(self, rays=None):
         if rays is None:
             self.rays = []
         else:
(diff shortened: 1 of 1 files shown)


GabGabG:
More modifications to come

GabGabG:
I removed more `if self.rays is not None` (or `is None`), because it was not really necessary (by default, `self.rays` is an empty list). It can be necessary if the user manually affect `Rays.rays = None`. If we want to prevent that, we can create a property and set `self.rays` to be an intern variable, like `self._rays`, this way it is more obvious that we should not touch this variable.

Also, I was wondering: in `Rays.append()`, we invalidate chached values, but since it can be time consuming to compute them, should it be better to append the new yValue dans thetaValue (the others would need to be re computed)?

raytracing/rays.py
```diff
@@ -99,46 +99,40 @@ def display(self, title="Intensity profile", showTheta=True):
...
     def displayProgress(self):
         nRays = len(self)
         if self.iteration % self.progressLog == 0:
             self.progressLog *= 3
             if self.progressLog > nRays:
                 self.progressLog = nRays
 
-            print("Progress {0}/{1} ({2:.0f}%) ".format(self.iteration, nRays, self.iteration/nRays*100))
-
-    def __len__(self) -> int:
```
dccote, line 130: I understand that this looks unnecessary : 
We want `Rays` to be iterable, and we only need to overload `__iter__` and `__next__`: https://docs.python.org/3/library/stdtypes.html#typeiter

However, `Rays` is the base class for `RandomRays`, which generate their rays on-demand up to a maximum `maxCount` and don't have an array with elements in it when created.  My goal was to have both Ray and Rays appear similar, which is why I added __len__ to Rays, even if it looks redundant: we can always do `len(Rays)` regardless of the class or subclass.

GabGabG: Ok I see, I will make the changes

GabGabG:
I re added the `if self.rays is None` and `if self.rays is not None` where they were.

PR #131, opened 2020-05-20
testsMaterials

MFournierQC:
Unit Tests for Materials.py with 100% tests coverage.


dccote:
To make sure we don't forget a new material, I suggest we use this:

```
for material in Materials.__subclasses__:
   # do test
   # make sure self.assert() provides information about which class failed
```

MFournierQC:
Your code suggestion didn't quite work for me but I found a different way of doing it that works pretty well ; 

```python
self.subclasses = []

for material in dir(materials):
   object = getattr(materials, material)
   try:
      if (object != Material) and issubclass(object, Material):
         self.subclasses.append(object)
      except TypeError:
         pass
```
This code has the advantage that it can find subclasses of subclasses of Material, should we end up needing that someday.

However, I do have a problem. Although the subclasses are pretty much in the same order every time I run my tests, there is no guarantee of where a new subclass would end up in the list. Knowing that, how would w go about testing the refractive index of each subclass respectively?

Creating some sort of list of pre defined refractive indexes doesn't seem like the ideal solution.

MFournierQC:
For testing actual good values, I went with this finally ;

```python
    def testMaterialSubclasses(self):
        refractiveIndexes = [1.3965252243506636, 1.5178527121226975, 1.5385861348077337, 1.5612286489801115,
                             1.539610715563772, 1.664053106820355, 1.5948478106562147, 1.6396821789377511,
                             1.5469302146171202, 1.5713583894047798, 1.4716376125966693, 1.459391121152617,
                             1.4251526811342752, 1.5810761652641074, 1.7197092324835102, 1.528065402814606,
                             1.541289278092723, 1.5975207997018837, 1.3404572894914368]

        for subclass in self.subclasses:
            n = subclass.n(5)
            if n not in refractiveIndexes:
                self.fails.append('Wrong value for subclass {}, {} not a valid n value.'.format(subclass.__name__, str(n)))
```
The thing is I really don't like this method but that seems the least complex method we could do. Let me know if you think you might have a better idea. Otherwise, I think I got something that automatically checks subclasses one by one and run the proper tests. So if you were to add a new subclass of material, the only thing you'd need to do for the test is to add the refractive index for a wavelength of 5 to the list.

MFournierQC:
Quick note also to say that, with the changes, the coverage is still 100%.

dccote:
> Your code suggestion didn't quite work for me but I found a different way of doing it that works pretty well ;
> 
> ```python
> self.subclasses = []
> 
> for material in dir(materials):
>    object = getattr(materials, material)
>    try:
>       if (object != Material) and issubclass(object, Material):
>          self.subclasses.append(object)
>       except TypeError:
>          pass
> ```
> 
> This code has the advantage that it can find subclasses of subclasses of Material, should we end up needing that someday.
> 
> However, I do have a problem. Although the subclasses are pretty much in the same order every time I run my tests, there is no guarantee of where a new subclass would end up in the list. Knowing that, how would w go about testing the refractive index of each subclass respectively?
> 
> Creating some sort of list of pre defined refractive indexes doesn't seem like the ideal solution.

I found the fix for subclasses, I think it is clearer: `__subclasses__()` is a function.  

For testing the values, I don't think the current test is good: first, we want to use visible or infrared wavelengths, because everything we use is in that range, so 5 µm is too far.  Second, too many digits in the current test: if we get that from the function, then we are not testing much.  We should look up the values elsewhere (the graph on refractive index.info) and use a value at say, 500 nm and use the value with a single digit after 1.x check if it is fine.

raytracing/tests/testsMaterials.py
```diff
@@ -0,0 +1,58 @@
...
+        ''' These are sample values of refractive indexes for each subclass of material for a wavelength of 5 micron.
+        In case of a new category of material, make sure to add the sample value to the list. '''
+        fails = []
+        refractiveIndexes = [1.3965252243506636, 1.5178527121226975, 1.5385861348077337, 1.5612286489801115,
+                             1.539610715563772, 1.664053106820355, 1.5948478106562147, 1.6396821789377511,
+                             1.5469302146171202, 1.5713583894047798, 1.4716376125966693, 1.459391121152617,
+                             1.4251526811342752, 1.5810761652641074, 1.7197092324835102, 1.528065402814606,
+                             1.541289278092723, 1.5975207997018837, 1.3404572894914368]
+
+        for subclass in self.subclasses:
```
dccote, line 47: For testing the values, I don't think the current test is good: first, we want to use visible or infrared wavelengths, because everything we use is in that range, so 5 µm is too far.  Second, too many digits in the current test: if we get that from the function, then we are not testing much.  We should look up the values elsewhere (the graph on refractive index.info) and use a value at say, 500 nm and use the value with a single digit after 1.x check if it is fine.  

Also, we should use a dictionary to match the class name with the value.

PR #135, opened 2020-05-21
Rays only accepts interables of Ray instances

### LINKED ISSUE #134: Rays() accepts objects from every type
This can cause further problems. In `__init__` itself it is not a big deal (except for coherence and logic) if `Rays` receive a string of nonsense or a list of numbers, but other methods will inevitably fail with exceptions of different types.

GabGabG:
`Rays.__init__()` was accepting every possible object and this would lead to *unexpected* outcomes in other methods. Now, it only accepts iterables (like a `list` or a *NumPy* `array`) and their content must be `Ray` instances. 
Fixes #134.

### DIFF
--- a/raytracing/rays.py
+++ b/raytracing/rays.py
@@ -20,10 +20,10 @@ class Rays:
             try:
                 inputOk = all([isinstance(ray, Ray) for ray in rays])
             except TypeError:  # The object is not iterable
-                raise AttributeError("'rays' parameter must be iterable (i.e. a list or a tuple).")
+                raise TypeError("'rays' parameter must be iterable (i.e. a list or a tuple).")
 
             if not inputOk:
-                raise AttributeError("'rays' parameter must contain Ray instances only.")
+                raise TypeError("'rays' parameter must contain Ray instances only.")
             self.rays = list(rays)
 
         self.iteration = 0
--- a/raytracing/rays.py
+++ b/raytracing/rays.py
@@ -167,6 +167,8 @@ class Rays:
         return self.rays[item]
 
     def append(self, ray):
+        if not isinstance(ray, Ray):
+            raise TypeError("'ray' must be a 'Ray' object.")
         if self.rays is not None:
             self.rays.append(ray)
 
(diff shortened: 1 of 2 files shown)


raytracing/tests/testsRays.py
```diff
@@ -7,6 +7,34 @@
...
+        rays = Rays(listOfRays)
+        self.assertListEqual(raysFromList.rays, listOfRays)
+        self.assertListEqual(raysFromTuple.rays, listOfRays)
+        self.assertListEqual(raysFromArray.rays, listOfRays)
+        self.assertListEqual(Rays(rays).rays, listOfRays)
+
+        with self.assertRaises(TypeError) as error:
+            # This should raise an exception
+            Rays("Ray(), Ray(1), Ray(1,1)")
+        self.assertEqual(str(error.exception), "'rays' parameter must contain Ray instances only.")
```
dccote: We don't check ths str() of an exception because they can change or be translated.
I fixed it already.

raytracing/rays.py
```diff
@@ -4,20 +4,36 @@
...
 """
 
-
 class Rays:
     def __init__(self, rays=None):
         if rays is None:
             self.rays = []
         else:
-            self.rays = rays
+            if isinstance(rays, collections.Iterable):
```
dccote, line 30: I rewrote this for clarity: I think it is better to rely of actual checks than exceptions.  It is clearer in this case.

raytracing/tests/testsRays.py
```diff
@@ -7,6 +7,31 @@
...
+
+        with self.assertRaises(TypeError) as error:
+            # This should raise an TypeError exception
+            Rays("Ray(), Ray(1), Ray(1,1)")
+
+        with self.assertRaises(TypeError) as error:
+            # This should raise an TypeError exception
+            Rays([Ray(), [1, 2], Ray()])
+
+        with self.assertRaises(TypeError) as error:
```
dccote, line 31: I don't think the AttributeError was correct also.

Issue #138, opened 2020-05-21
Example section 9 : Widefield-microscope


Valeriouus:
When the lenses in a 4f system are too small, vignetting. We show a simple imaging system with vignetting, Field Stop poorly placed at a lens instead of the image with fix in next figure + Aperture stop. 
- Make a code producing a scheme of a simple 4f system with lenses too small = vignetting ! 
- Modify this code to produce a scheme of the same 4f system with bigger lenses = no more vignetting and good image size

PR #139, opened 2020-05-21
Issue75 example transmission

Valeriouus:
Put more people to review than less. Feel free to comment (or not) according to your mood.


dccote:
Good but we need to use the magnification of the system instead of a hard number for the ideal
Pinhole size. I’m thinking about how to do that.

shadimasoumi:
> Good but we need to use the magnification of the system instead of a hard number for the ideal
> Pinhole size. I’m thinking about how to do that.

Maybe we can use a defined object with a special size and compare the result for different pinhole sizes, do you think does that work?

dccote:
> Maybe we can use a defined object with a special size and compare the result for different pinhole sizes, do you think does that work?

We are defining the size of the focal spot with 0.000250. in the rays, so I think we should use than and `magnification()` but the code will be confusing we need to have from focus to pinhole, then do the loop

I think the simplest would be to state at the beginning that the magnification is -18.75.


```
from raytracing import *
import matplotlib.pyplot as plt

illumination = ImagingPath()
illumination.append(Space(d=5))
illumination.append(Lens(f=5))
illumination.append(Space(d=105))
illumination.append(Lens(f=100))
illumination.append(Space(d=200))
illumination.append(Lens(f=100))
illumination.append(Space(d=175))
illumination.append(Lens(f=75))
illumination.append(Space(d=115))
illumination.append(Lens(f=40))
illumination.append(Space(d=90))
illumination.append(Lens(f=50))
illumination.append(Space(d=50))
(Mt, Ma) = illumination.magnification()
print("Magnification: {0}".format(Mt))
```

Valeriouus:
I'm not sure there is a function magnification (if there is, I can't find it), but there is a function called imageSize() in imagingpath.py. I used it this was : 
illumination = ImagingPath()
        illumination.append(Space(d=newPositions))
        illumination.append(Lens(f=5))
        illumination.append(Space(d=105))
        illumination.append(Lens(f=100))
        illumination.append(Space(d=200))
        illumination.append(Lens(f=100))
        illumination.append(Space(d=175))
        illumination.append(Lens(f=75))
        illumination.append(Space(d=115))
        illumination.append(Lens(f=40))
        illumination.append(Space(d=90))
        illumination.append(Lens(f=50))
        illumination.append(Space(d=50))
        pinholeSize = illumination.imageSize()*pinhole # Ideal size of the pinhole times a certain factor
        illumination.append(Aperture(diameter=pinholeSize))

However, it prints in the terminal kind of a matrix because of this line "print(distance, conjugateMatrix)", which is perhaps a bit annoying for the user. Can we take this line out in some way? Or delete it ?

JLBegin:
Is this the expected output ?
image

JLBegin:
> However, it prints in the terminal kind of a matrix because of this line "print(distance, conjugateMatrix)", which is perhaps a bit annoying for the user. Can we take this line out in some way? Or delete it ?

`ImagingPath.imageSize()`
Indeed, I don't think a method is supposed to print anything by default unless the title explicitly say that it will print something... @dccote ?

dccote:
> print(distance

Done. This was left over debugging code.

Valeriouus:
We have the same problem for widefield.py, I don't know why. #138
Screen Shot 2020-05-25 at 10 03 01 AM

JLBegin:
> We have the same problem for widefield.py
It's not the same branch. Widefield is now on master while the fix was only made on the branch for this issue75. This will be solved once we merge this PullRequest.

PR #146, opened 2020-05-22
RandomRays.rays is always None and RandomRays.__getitem__ always raises an exception

### LINKED ISSUE #145: RandomRays should always be inaccessible with __getitem__ and its rays None
For a normal `Rays`object, it makes sense to access a specific ray, but for a `RandomRays`, it does not:
 - They are created *on the fly*, so they are not stored nor created in advance

It also makes more sense to have `None` as its `rays` properties, since it does not have any ray, never. It creates `Ray` objects, but it does not keep them, so `None` seems to make more sense than `[]`, like it is right now.

I suggest two changes, coming in a future pull request:
- We explicitly set its `rays` property as `None` (now: it is `[]`)
- `__getitem__` always raises an exception (now: it sometimes does, but sometimes not, which can lead to a bigger problem: index out of bounds, because `rays` property is always an empty list)

GabGabG:
Since `RandomRays` creates `Ray` objects on demand, it doen't make sense to have `RandomRays.rays = []`. I think it should always be `None`.

Since `RandomRays` creates `Ray` objects on demand, it should never be accessed with `__getitem__`, because it doesn't contain any `Ray`. It is also a problem if we keep `rays = []`.
Fixes #145

### DIFF
--- a/raytracing/rays.py
+++ b/raytracing/rays.py
@@ -22,6 +22,7 @@ or when analysing the resulting rays that reached a plane in ImagingPath,
 MatrixGroup or any tracing function. 
 """
 
+
 class Rays:
     def __init__(self, rays=None):
         if rays is None:
@@ -232,6 +233,7 @@ class Rays:
     # https://en.wikipedia.org/wiki/Xiaolin_Wu's_line_algorithm
     # and https://stackoverflow.com/questions/3122049/drawing-an-anti-aliased-line-with-thepython-imaging-library
 
+
 class UniformRays(Rays):
     def __init__(self, yMax=1.0, yMin=None, thetaMax=pi / 2, thetaMin=None, M=100, N=100):
         self.yMax = yMax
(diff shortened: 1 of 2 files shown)


raytracing/rays.py
```diff
@@ -22,6 +22,7 @@
 MatrixGroup or any tracing function. 
 """
 
+
 class Rays:
     def __init__(self, rays=None):
         if rays is None:
```
dccote, line 28: Rays as a base class wants an array _rays, but it should not be writable.
RandomRays will create rays as it goes because if not, it is too lengthy. It will append them to the array as they are created.

raytracing/rays.py
```diff
@@ -284,18 +286,14 @@ def __init__(self, yMax=1.0, yMin=None, thetaMax=pi / 2, thetaMin=None, maxCount
         self.thetaMin = thetaMin
         if thetaMin is None:
             self.thetaMin = -thetaMax
-        super(RandomRays, self).__init__(rays=None)
+        super(RandomRays, self).__init__()  # Even with this, rays = []
```
dccote, line 289: Either way, RandomRays need to provide a protected access to __getitem__ without exposing _rays.
If rays have already been produced, take them from array, if they have. not, generate up to the item and return.

raytracing/rays.py
```diff
@@ -284,18 +286,14 @@ def __init__(self, yMax=1.0, yMin=None, thetaMax=pi / 2, thetaMin=None, maxCount
...
         return self.maxCount
 
     def __getitem__(self, item):
-        if self.rays is None:
-            raise NotImplementedError("You cannot access RandomRays")
-        elif len(self.rays) < item:
-            raise NotImplementedError("You cannot access RandomRays")
-        else:
-            return self.rays[item]
+        raise NotImplementedError("You cannot access RandomRays")
```
dccote, line 296: something like:

```
while len(self.rays) < item:
    # generate and store in self.rays

return self.rays[item]

File: raytracing/rays.py [hunk 4]
```diff
@@ -284,18 +286,14 @@ def __init__(self, yMax=1.0, yMin=None, thetaMax=pi / 2, thetaMin=None, maxCount
...
 
     def __next__(self) -> Ray:
         if self.iteration >= self.maxCount:
```
dccote, line 299: This will need to be modified too to make sure the behaviour is correct: generate next if not already there, if not, use self.rays[self.iteration]

PR #150, opened 2020-05-24
Example widefield microscope #138

shadimasoumi:


shadimasoumi:
> Very well done.
> I changed the language a bit and I added a line to only show the principal and axial rays. Let me know what you think: do you think it is better to show the full object or the "Object" as obtained from the field of view?

Very good! This version is better.
I think the example is easier to understand now!

PR #176, opened 2020-05-26
Load check if the content of the loaded file is iterable and contains only Ray instances

### LINKED ISSUE #175: Rays can load any "pickled" file
This can lead to unwanted behaviors. We should check if the content of the file is an iterable of `Ray` objects.

`load` directly extends (or override) the inner list or rays in `Rays`. This can cause unexpected problems without proper guidance to what caused the issue.

GabGabG:
Not checking this can lead to unexpected behaviors in the code.

Fixes #175

### DIFF
--- a/raytracing/tests/testsRays.py
+++ b/raytracing/tests/testsRays.py
@@ -69,7 +69,6 @@ class TestRays(unittest.TestCase):
         r = Rays(listOfRays)
         self.assertEqual(len(r), len(listOfRays))
 
-
     def testRaysGetRay(self):
         raysList = [Ray(), Ray(1), Ray(-1)]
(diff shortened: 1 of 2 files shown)


raytracing/tests/testsRays.py
```diff
@@ -258,7 +256,7 @@ def setUp(self) -> None:
         self.fileName = 'testFile.pkl'
         with open(self.fileName, 'wb') as file:
             pickle.Pickler(file).dump(self.testRays.rays)
-        time.sleep(1)  # Make sure everything is ok
```
dccote, line 261: On my computer I needed 1.0 second.  Not sure exactly what was going on.

PR #177, opened 2020-05-26
Remove useless part in traceManyThrough

GabGabG:
We don't need to check if the input is an iterable of `Ray` since `Rays.__init__` does it. This is just a little update.


dccote:
Technically, `Rays` came after the "list of rays".  I don't know if we need to keep the option to use list of rays or if we will always use Rays.

PR #196, opened 2020-05-29
ImageSize of ImagingPath returns inf when there is no forward conjugate

### LINKED ISSUE #188: ImagingPath: imageSize fails when D is 0
When `D == 0`, there is no forward conjugate (nor backward), so we return `conjugateMatrix` as `None`. Because of that, `imageSize` raises an `AttributeError`. We should check if `conjugateMatrix` is `None`, but what should we return? Since there is no image, return `None`?

GabGabG:
There was a problem with `imageSize` when the current `ImagingPath` has `D == 0`, it raised an `AttributeError` because the forward conjugate matrix doesn't exist (it is `None`). The code now checks `if conjugateMatrix is None` then returns `float("+inf")` if it is the case.

Fixes #188

### DIFF
--- a/raytracing/imagingpath.py
+++ b/raytracing/imagingpath.py
@@ -585,6 +585,9 @@ class ImagingPath(MatrixGroup):
         """
         fieldOfView = self.fieldOfView()
         (distance, conjugateMatrix) = self.forwardConjugate()
+        if conjugateMatrix is None:
+            return float("+inf")
+
         magnification = conjugateMatrix.A
         return abs(fieldOfView * magnification)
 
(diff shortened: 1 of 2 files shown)


raytracing/imagingpath.py
```diff
@@ -585,6 +585,9 @@ def imageSize(self):
         """
         fieldOfView = self.fieldOfView()
         (distance, conjugateMatrix) = self.forwardConjugate()
```
dccote, line 587: Should we check distance == +inf or matrix is None?

dccote: I am also torn between Matrix == None or a matrix with infinite element B. Not sure what to do here or what it implies to have a matrix with +inf in the calculations.

PR #207, opened 2020-05-29
Proposition: displayRange deals in absolute values

GabGabG:
We should consider the absolute value of the object height and the magnification.
Consider this example:
```python
path = ImagingPath(System4f(10, 10, 20, 20))
path.append(Space(5))
path.append(CurvedMirror(-14, 20))
path.objectHeight = 20
path.display()
```
We get:
figPullRequest

And if we consider absolute values, we get:
fig2PullRequest

I think it would be great to consider the absolute values, because it seems more coherent. Why should we consider positive magnification only?


dccote:
Ok, I am trying to see exactly what was the problem that you solved (!) but the reason magnification is always signed is that it represents inversion: if it is negative, the image is inverted.  However, for calculations, it makes sense to take only the absolute value.  But why was objectHeight negative sometimes?

GabGabG:
> But why was objectHeight negative sometimes?

I did that in case a user set it so. It is a possible outcome in the current code and I thought it would be important to consider it.

raytracing/imagingpath.py
```diff
@@ -693,14 +693,16 @@ def displayRange(self, axes=None):
         """
 
         displayRange = self.largestDiameter
-        if displayRange == float('+Inf') or displayRange <= 2 * self.objectHeight:
-            displayRange = 2 * self.objectHeight
+        objHeight = abs(self.objectHeight)
```
dccote, line 696: If it is never valid, then we should create a @property getter and setter to prevent using a negative value in the first place.

raytracing/imagingpath.py
```diff
@@ -693,14 +693,16 @@ def displayRange(self, axes=None):
...
+        if displayRange == float('+Inf') or displayRange <= 2 * objHeight:
+            displayRange = 2 * objHeight
 
         conjugates = self.intermediateConjugates()
         if len(conjugates) != 0:
             for (planePosition, magnification) in conjugates:
-                if displayRange < self.objectHeight * magnification:
```
dccote, line 702: then we should take absMagnification = abs(magnification) for this part.

GabGabG:
I updated the code to make `self.objectHeight` a property and created a new method `setObjectHeight` to set a new value. If the new value is negative, a `ValueError` is raised. I updated some code that set `self.objectHeight = something`. I added documentation to setter/getter (although I'm not sure it is enough or if there is something missing in the doc).

JLBegin:
Oh. Change of API... Can't we make an objectHeight property and define its setter with @objectHeight.setter instead? Or it has to make a call even in this case?

dccote:
> Oh. Change of API... Can't we make an objectHeight property and define its setter with @objectHeight.setter instead? Or it has to make a call even in this case? Sorry I'm afk

I agree, and I just did.

raytracing/imagingpath.py
```diff
@@ -102,6 +102,20 @@ def __init__(self, elements=None, label=""):
         self.showPlanesAcrossPointsOfInterest = True
         super(ImagingPath, self).__init__(elements=elements, label=label)
 
+    @property
```
dccote, line 105: @GabGabG Please take a look at this: @property getter and setter allow for `path.objectHeight = 20`.

GabGabG:
I didn't know we could use such a decorator to set a property! Alright, this looks fine and clean, I fixed some code I modified earlier that still had `set...`.

PR #209, opened 2020-05-30
Documentation rays.py

ElaheParham:
The documentation for the following is updated:
ray.py
rays.py

The names of some contributors are added to the first page. (Others will be added in the next update because I didn't know their complete names :) )

The Contribute page (in the documentation) now includes the procedure for bug reporting.


dccote:
I updated the list.

PR #213, opened 2020-05-31
issue#120_NewSpecialtyLensType

MFournierQC:


MFournierQC:
Implemented a new class, AchromatSingletLens (name can be changed) for some lenses that weren't handled very well.

From testing, I get no UserWarnings from the new class. I also checked for ImagePath and the lenses seem to be displayed correctly.

I think the name could probably be changed to simply Lens and the class could be moved out of SpecialtyLenses though? To discuss.

raytracing/specialtylenses.py
```diff
@@ -205,6 +205,156 @@ def pointsOfInterest(self, z):
         return [{'z': f1, 'label': '$F_f$'}, {'z': f2, 'label': '$F_b$'}]
 
 
+class SingletLens(MatrixGroup):
```
dccote, line 208: A Singlet lens cannot be achromatic: the index of refraction is never constant, and obviously the radii of curvature will not change.  Therefore, the lens has to suffer from chromatic aberrations, that's why it is called just 'Singlet'.  The AchromaticDoublet solve that problem partially with two materials.

raytracing/specialtylenses.py
```diff
@@ -205,6 +205,156 @@ def pointsOfInterest(self, z):
...
+    """
+        General singlet lens with an effective focal length of f, back focal
```
dccote, line 210: We use the same nomenclature as Thorlabs for now. The figure is available on the product page:
https://www.thorlabs.com/images/TabImages/Plano-Convex_Lens_Schematic.gif

raytracing/eo.py
```diff
@@ -12,7 +12,7 @@ class PN_33_922(AchromatDoubletLens):
     def __init__(self):
         # PN for Part number
         super(PN_33_922,self).__init__(fa=150.00,fb=126.46, R1=92.05,R2=-72.85, R3=-305.87, 
-                                    tc1=23.2, tc2=23.1, te=36.01, n1=1.6700, n2=1.8467, diameter=75,
+                                    tc1=23.2, tc2=23.1, te=36.01, n1=N_BAK1.n(0.5876), n2=N_SF8.n(0.5876), diameter=75,
```
dccote, line 15: I found the problem on Edmund Optics web site, see commit https://github.com/DCC-Lab/RayTracing/commit/d026dabe2f600cb9ec4c39266814acbb73e3a9d3

raytracing/specialtylenses.py
dccote: Confirm you have seen this, then I will merge.

raytracing/thorlabs.py
```diff
@@ -32,7 +32,15 @@ def __init__(self):
 
 class AC254_030_A(AchromatDoubletLens):
     def __init__(self):
-        super(AC254_030_A,self).__init__(fa=30.0,fb=22.9, R1=20.89, R2=-16.7, R3=-79.8,
+        """ As of May 31 2020, there is an error on the product web page at:
```
dccote, line 35: Inspired by my finding on Edmund Optics web site, I found this mistake on Thorlabs' web site.

raytracing/thorlabs.py
```diff
@@ -96,6 +104,11 @@ def __init__(self):
 
 class AC254_125_A(AchromatDoubletLens):
     def __init__(self):
+        """ As of May 31 2020, the calculated edge thickness does not match
```
dccote, line 107: Although I am convinced it is an error, I am unable to confirm this number.  Let's leave it like that for now.

raytracing/specialtylenses.py
MFournierQC: I already knew that but it slipped my mind. Indeed, SingletLens makes much more sense than AchromaticSingletLens.

raytracing/eo.py
MFournierQC: I thought the composing glasses were N-BAF10 and N-SF57? When I looked up for those, I found the same n1 and n2 as the values before the change. I'd be interested to know where you found the new values?

dccote: As mentionned in the commit https://github.com/DCC-Lab/RayTracing/commit/d026dabe2f600cb9ec4c39266814acbb73e3a9d3, this file is the actual "design reference": 
https://www.edmundoptics.com/document/download/405650
Note 1, at the top left, says the actual material used in the modelling.

PR #215, opened 2020-06-01
New FigureManager and Drawing API

JLBegin:
Still in development.

### Main changes
*todo*


raytracing/drawing.py
```diff
@@ -0,0 +1,324 @@
+import matplotlib.pyplot as plt
+from matplotlib import patches, transforms
+from matplotlib import text as mplText
+import matplotlib.path as mpath
+from typing import List
+import numpy as np
+import math
+
+
+class Drawing:
```
dccote, line 10: I think the class should be called `Graphic` or `Graphics`

JLBegin: Graphic yes, it makes sense

raytracing/drawing.py
```diff
@@ -0,0 +1,324 @@
...
+        if label is not None:
+            self.label = Label(text=label, x=self.centroid[0], y=self.halfHeight() * 1.2)
+
+    @property
+    def hasLabel(self) -> bool:
+        if self.label is None:
+            return False
+        return True
+
+    def applyTo(self, axes: plt.Axes, x: float = None, y: float = None):
```
dccote, line 62: `Drawing` should be GUI-agnostic: I would not use matplotlib terminology because we could change it later. You could define your own (x,y) coordinates, which would be in dataspace anyway(?) so regardless of the choice of graphics backend, you could use your class.

raytracing/drawing.py
```diff
@@ -0,0 +1,324 @@
...
+            return 1, 1
+
+        xScale, yScale = self.axesToDataScale()
+
+        heightFactor = self.halfHeight() * 2 / yScale
+        xScaling = xScale * (heightFactor / 0.2) ** (3 / 4)
+
+        return xScaling, 1
+
+    def axesToDataScale(self):
```
dccote, line 124: This would be overridden in a `MatplotlibGraphics`, `CocoaGraphics`, `TkGraphics`, etc... Everything else does not need to know the details of the backend.

raytracing/drawing.py
```diff
@@ -0,0 +1,324 @@
...
+    @property
+    def centroid(self):
+        xy = []
+        for component in self.components:
+            xy.extend(component.get_xy())
+
+        return np.mean(xy, axis=0)
+
+
+class SurfacePairPatch(patches.PathPatch):
```
dccote, line 149: You should then define `SurfacePair`,`Arrow`, `Aperture` and `Label` in general. They can manage most details except the actual drawing. Then you would have two options:
1. specialize them immediately with `SurfacePairPatch`, `ArrowPatch` etc... and draw with matplotlib
or
2. create a `Bezier` class, because that is the most general drawing in 2D: its interface would be general, because a Bezier curve is general. You could then create a `BezierPatch` or an internal variable that would implement the details in matplotlib, or any other backend.

I prefer option 2, it is much less work I think: a new backend would simply mean implementing Bezier.

JLBegin: I understand. Option 2 still requires to break down the arrow, doubleArrow, aperture, surfacePair, and future graphics in a list of bezier curves which takes some time.

dccote: Maybe you misunderstand what I am saying because I don't think it is that much work.
You could also have an option to add arrows at both ends of your Bezier curve.

dccote:
Temporarily closed to focus on smaller redesign.

PR #225, opened 2020-06-03
MatrixGroup: save setup

GabGabG:
I propose that we can save a `MatrixGroup` instance, like we do with `Rays`. I think it can be an interesting feature. That way, users can share setups by just sharing a `pickled` file, then load it and ***bam*** the setup is created without too much coding.

It can be useful for research, like sending files to other colleagues. It can also be useful for teaching purposes: instead of sending the script that builds the setup, you can only send the file with the setup saved.

It works like `Rays` saving/loading methods: we use `pickle` to save every components of the group and then we iterate on every components saved in the file to get the *original* setup. It keeps the original type of the element. For instance, it knows if the original component was a `Lens` or `ThickLens`, it does not only save the ABCD matrix.

It is fully tested.

I think it is a key enhancement.

### DIFF
--- a/raytracing/matrixgroup.py
+++ b/raytracing/matrixgroup.py
@@ -236,7 +236,16 @@ class MatrixGroup(Matrix):
             return element
         raise StopIteration
 
-    def saveElements(self, filePath: str):
+    def save(self, filePath:str):
+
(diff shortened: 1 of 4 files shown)


GabGabG:
Documentation is missing for now

PR #226, opened 2020-06-03
Allow for OnlyAxialRays and new design parameters handler

JLBegin:
Added **new design parameters** by Valerie's request. 
- `onlyAxialRays` : only show the rays coming from the center of the object.
- `imageColor` : color used for the arrow of all images. 
- `objectColor`

The design parameters can now be updated with a **new method** like so:
```
path = ImagingPath()
path.design(rayColors=['r', 'r', 'r'], onlyAxialRays=True)
```
This adds the benefit of autocompletion since the design method holds all design parameters. 
It doesn't really change the API yet since rayColors was not being used.

### DIFF
--- a/raytracing/imagingpath.py
+++ b/raytracing/imagingpath.py
@@ -87,7 +87,7 @@ class ImagingPath(MatrixGroup):
         self.fanAngle = 0.1  # full fan angle for rays
         self.fanNumber = 9  # number of rays in fan
         self.rayNumber = 3  # number of points on object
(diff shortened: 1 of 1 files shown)


dccote:
I should have merged this before the redesign of the display class.  It is not a bad idea, but now, with `ImagingPath` having lost all drawing commands, it cannot be merged intelligently.  Let's move these to `Figure` by starting over from master.

dccote:
And there is a single axial ray, while there are many marginal rays.

PR #232, opened 2020-06-03
Intermediate redesign of display

JLBegin:
Removed all draw-related methods from ImagingPath and Matrix Objects.
Display methods are now in a Figure class which calls the required MatrixGraphics for each element in the path (same old draw() methods though)


JLBegin:
I fixed unittests for Matrix, MatrixGroup and ImagingPath. Also added testsFigure to hold previously written tests for displayRange and axesToDataScale.

PR #236, opened 2020-06-05
Get, set and other new methods in MatrixGroup

GabGabG:
I implemented new methods in  `MatrixGroup`: `__len__` to easily get the number of matrices, `__getitem__` to either get a single `Matrix` if we give an integer or a `MatrixGroup` if we give a `slice`, `removeElement` to remove a `Matrix` at a specific index and if we want to pad with `Space` or *squeeze* everything, `insertElement` which inserts a single `Matrix` or a `MatrixGroup` before an element at a specific index, `replaceSingle` which replaces a single element (integer index) with a `Matrix` or `MatrixGroup`, `replaceChunk` to replace more than one element (`slice` index) with a `Matrix` or `MatrixGroup`. Finally, `__setitem__` is implemented and depending on the key (integer or `slice`), we use `replaceSingle` or `replaceChunk`.

I think it can be an interesting implementation, because it allows the option to directly modify the current `MatrixGroup`, `LaserPath`, `ImagingPath`, etc. It is useful in cases where we have a specific setup and we want to see what happens when we change some components. We don't need to copy-paste the code creating the group and create a new one. For example:
```python
mg = MatrixGroup([Space(10), Lens(10), Space(20), Lens(10), Space(10), CurvedMirror(-20)])
<code here>
mg[-1] = CurvedMirror(-15) # Change the mirror at the end
<code here>
mg.removeElement(-1) # Remove the mirror, only 4f system
<code here>
```

100% tested. I think it is a great enhancement.


raytracing/matrixgroup.py
```diff
@@ -235,3 +235,119 @@ def __next__(self):
...
+
+    def __len__(self):
+        return len(self.elements)
+
+    def __getitem__(self, item):
+        if isinstance(item, slice):  # If we get a slice, return a matrix group
+            return MatrixGroup(self.elements[item])
+        return self.elements[item]
+
+    def removeElement(self, index: int, pad: bool = False):
```
dccote, line 247: We use `append()`, should we use `remove()` and `insert()` ? It is not coherent to have both I think.

raytracing/matrixgroup.py
```diff
@@ -235,3 +235,119 @@ def __next__(self):
...
+            if i == index:
+                toReplace = tempElements[i]
+                if toReplace.L != newElement.L:
+                    warnings.warn("Physical length mismatch. Squeezing or extending the current group.", UserWarning)
+                for element in newElement:
+                    self.append(element)
+            else:
+                self.append(tempElements[i])
+
+    def replaceChunk(self, startIndex: int, stopIndex: int, newChunk: Matrix):
```
dccote, line 312: I don't like this name. Chunk. Group?

raytracing/matrixgroup.py
```diff
@@ -235,3 +235,119 @@ def __next__(self):
...
+            for i in range(maxIndex):
+                previousElement = tempElements[i]
+                if i == index:
+                    for newElement in element:
+                        self.append(newElement)
+                    self.append(previousElement)
+                else:
+                    self.append(previousElement)
+
+    def replaceSingle(self, index: int, newElement: Matrix):
```
dccote, line 290: Can't this be done with `__getitem__` ?

GabGabG: No because we need `__setitem__` to support assignment. That's why I thought of ways to support it. We can't just do
```python
mg = MatrixGroup([Space(10), Lens(5), Space(10)])
mg[1] = Lens(10)
```
with only `__getitem__`.

GabGabG:
@dccote this was mostly a draft with *brute force* code, not very optimized at first glance. I don't really see how it makes the API complex to be honest. I fear this may seem useless, but I think it can help. Like I wrote, this allows for simple inplace changes in a group without instancitating a new object. I see an example where it could be useful: you can build a quite complex setup with objectives and mirrors, but then you want to try with different lenses or apertures to see the impact on the final image (or wathever parameter you are intersted in). You simply have to access the group and change the initial components of the group without the need of a new instance or going back in the code to change the initial append/initialization. I don't have a *coded* example where we have said setup, but some kind of complex microscope can be one.

The comments on `append` and `insert` is interesting, I didn't think of that in the first place. The code can be changed without too many problems.

If you think this is not useful, it doesn't bother me to give up this idea.

dccote:
I like the idea, but I'd like the API to be simpler. If we only had  `insert`, `remove` and used `__getitem__`? Should we have essentially the Python list-API? I don't know.

Another option: is there an "automatic" way of forwarding all *list* commands to self.elements?

GabGabG:
I'll try to simplify some things. I think it is possible to have a code less *complicated* and a better API.

GabGabG:
> I like the idea, but I'd like the API to be simpler. If we only had `insert`, `remove` and used `__getitem__`? Should we have essentially the Python list-API? I don't know.

For the list API, it is possible but it adds a lot of methods (count, clear, copy, and all `__dunder__` methods). Only implementing `insert`, `remove` and `_getitem__` seems good to me.

> Another option: is there an "automatic" way of forwarding all _list_ commands to self.elements?

It is possible to redirect methods to an attribute (see https://stackoverflow.com/questions/13460889/how-to-redirect-all-methods-of-a-contained-class-in-python/41040380) I tried a little basic mini test:
```python
class Toto:
    def __init__(self):
        self.list = []

    def __getattr__(self, item):
        if not item in self.__dict__: # We check if the item is not an attribute/method of the current class
            return getattr(self.list, item) # If so, we return self.list.item(...)
        return self.__dict__[item] # Else, we return self.item(...)

if __name__ == "__main__":
    t = Toto()
    t.append(1) # Since Toto does not have any method called append, we use list.append
    print(t.list)
```
The output is:
```python
[1] # At initialization, the list was empty. We appended 1 through __getattr__
```
It can be useful if we don't want too many methods in the API, but I see an issue (that can be fixed with a method I think): when we append, insert, delete, etc. we need to check if there is a mismatch between indices of refraction (and compute ABCD). Using custom methods like I wrote prevent such issue, but using the base methods from `list` does not. What we can do is: 
1. Forward methods to `self.elements`
2. Use a method, maybe called `update` that check if everything fits and compute what needs to be computed

dccote:
Ok, I think the best is to only implement insert, remove and __getitem__ __setitem__, try to make the code as simple as possible.  Also, keep the methods together.

PR #246, opened 2020-06-08
New design parameters

JLBegin:
Inside Figure class there's a designParams dictionary and a design() method to easily update any of the parameter. 

**Changes**
- Moved `ImagingPath.rayTraceLines()` and `.rearrangeRayTracePlot()` inside Figure instead since it is plt/display related. 
- New design() method to easily modify the parameters (ImagingPath.design relays to Figure.design)
- Now creating an empty Figure object at ImagingPath.__init__ to pass user design calls before display() is called.
- Updated tests

**Parameters** available for now: `onlyAxialRay`, `rayColors`, `imageColor`, `objectColor`.


raytracing/figure.py
```diff
@@ -78,6 +80,32 @@ def addFigureInfo(self, text):
         self.axes.text(0.05, 0.15, text, transform=self.axes.transAxes,
                        fontsize=12, verticalalignment='top', clip_box=self.axes.bbox, clip_on=True)
 
+    def design(self, rayColors: List[Union[str, tuple]] = None, onlyAxialRay: bool = None,
```
dccote, line 83: I like the idea of a `design`.  We should have ready-made groups of parameters wrapped into a design, such as `presentation`, `publication`, `screen`, etc... For instance, publication would avoid colours if possible, no title, whereas presentation could have colours and larger fonts, etc...

raytracing/figure.py
```diff
@@ -78,6 +80,32 @@ def addFigureInfo(self, text):
...
+        onlyAxialRay : bool, optional
+            Only draw the ray fans coming from the center of the object (axial rays).
+            Works with fanAngle and fanNumber. Default to False.
+        imageColor : Union[str, tuple], optional
+            Color of image arrows. Default to 'r'.
+        objectColor : Union[str, tuple], optional
+            Color of object arrow. Default to 'b'.
+        """
+
+        params = locals()
```
dccote, line 101: I am not sure I like this hack. `locals()` has *all* local variables. It looks like a disaster waiting to happen.

raytracing/figure.py
```diff
@@ -78,6 +80,32 @@ def addFigureInfo(self, text):
...
+        for key, value in params.items():
+            if value is not None:
+                if key is 'rayColors':
+                    assert len(value) is 3, \
+                        "rayColors has to be a list with 3 elements."
+                self.designParams[key] = value
+
     def display(self, limitObjectToFieldOfView=False, onlyChiefAndMarginalRays=False,
```
dccote, line 109: we will need to change `onlyChiefAndMarginalRays` to something else because we have officially changed the nomenclature: these two rays are called the principal and axial rays. It would be nice to have something that is backward compatible.

dccote: I would also like the default to be limited to field of view and only principal and axial rays whenever possible, by default.

JLBegin:
Added styles (group of design parameters). You can set a style with .design('publication'). I try to avoid setters for now... what do you think ?

JLBegin:
> Also, maybe create functions to get groups of rays?

I'm not sure I understand. Do you mean something like having methods axialRayLines() and principalRayLines() that would get called by rayTraceLines ?

> I would also like the default to be limited to field of view and only principal and axial rays whenever possible, by default.

Do you want to try `limitObjectToFieldOfView=True` and automatically switch to False if it cannot? (right now it raises an error, I guess we could just show a warning instead).

dccote:
> > Also, maybe create functions to get groups of rays?
> 
> I'm not sure I understand. Do you mean something like having methods axialRayLines() and principalRayLines() that would get called by rayTraceLines ?
> 
> > I would also like the default to be limited to field of view and only principal and axial rays whenever possible, by default.
> 
> Do you want to try `limitObjectToFieldOfView=True` and automatically switch to False if it cannot? (right now it raises an error, I guess we could just show a warning instead).
> Same question with `onlyPrincipalAndAxialRays=True`.

Yes: change to False if not able to do it.

dccote:
The thin lens arrows are too large and have a different width depending on their sizes.  Can this be fixed?
image

JLBegin:
> The thin lens arrows are too thick and have a different width depending on their sizes. Can this be fixed: thinner and same size for every lens.
> image

Yes so we just scale the length of the arrow and arrow head ? (we still need some scaling down so that tiny lenses are drawn 'properly')

JLBegin:
Not scaling down the width leaves some problems with small lenses. I could adjust the relative arrow head size to make it bigger but I'm at risk of overlapping the arrow heads.
image

Maybe we can wait for the graphic update, since I will be using a different arrow style for thin lenses : 
image

PR #261, opened 2020-06-11
LaserPath separated into LaserPath and LaserCavity

dccote:
Added some tests at the same time and found crashers in GaussianBeam. This has been fixed, tested, and modified.


dccote:
We need to move the `display()` and `draw()` commands of `LaserPath` to `Figure`.
I started, but you should finish @JLBegin

JLBegin:
I think it works now. I have a problem with circular imports though. But anyway, here's an example of the display (not sure I am properly using the LaserPath API):
```python
from raytracing import *

beam1 = GaussianBeam(w=1)
beam2 = GaussianBeam(w=2)

path = LaserPath()
path.append(Space(d=10))
path.append(Lens(f=5))
path.append(Space(d=7))

path.display(inputBeams=[beam1, beam2])
```
image

dccote:
You need to get more points in space: don't draw only at the elements.  Get 20 points in between.

JLBegin:
I'm not sure I understand what you are saying. I went and run the old display code and I get a similar plot. 
I changed my test code so we can see some diffraction. Here is for one gaussian beam (w=0.01, R=10) : 

```python
path = LaserPath()
path.append(Space(d=10))
path.append(Lens(f=5, diameter=0.5))
path.append(Space(d=10))
path.display(inputBeams=[GaussianBeam(w=0.01, R=10)])
```
image

dccote:
> I'm not sure I understand what you are saying. I went and run the old display code and I get a similar plot.
> I changed my test code so we can see some diffraction. Here is for one gaussian beam (w=0.01, R=10) :
> 
> ```python
> path = LaserPath()
> path.append(Space(d=10))
> path.append(Lens(f=5, diameter=0.5))
> path.append(Space(d=10))
> path.display(inputBeams=[GaussianBeam(w=0.01, R=10)])
> ```
> 
> image

You are right, I also tested it and it's just that the Rayleigh range was large.  The code works fine.

PR #262, opened 2020-06-11
Added documentation to new methods (pop, insert, etc) + minor modifications

GabGabG:
I wrote documentation for the new methods I created. I also changed some parameter names, because they were different from one method to another even if their roles were very similar.

### DIFF
--- a/raytracing/matrixgroup.py
+++ b/raytracing/matrixgroup.py
@@ -98,18 +98,20 @@ class MatrixGroup(Matrix):
         return len(self.elements)
 
     def __getitem__(self, key):
-        """ This function is used to substitutes a single matrix or a whole group with a single or multiple matrices.
+        """ This function is used to obtain a single element.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


raytracing/matrixgroup.py
```diff
@@ -92,14 +92,88 @@ def append(self, matrix):
...
+        Returns the number of matrices in the group. Allows the use of len(MatrixGroup).
+        """
         return len(self.elements)
 
-    def __getitem__(self, item):
-        if isinstance(item, slice):  # If we get a slice, return a matrix group
-            return MatrixGroup(self.elements[item])
-        return self.elements[item]
+    def __getitem__(self, key):
+        """ This function is used to obtain a single element.
```
dccote, line 101: This used to say "substitute" an element.  I corrected it, is it only for getting an element?

GabGabG: Yes it is only for getting, it was a copy paster error

PR #265, opened 2020-06-11
Fixed Figure to use the correct nomenclature for principalRay and axial

dccote:
Regardless of limiting field of view  or not in display(), the princical ray is still the chief ray of the field of view limits.

There was a mistake in the principalRay calculation when chiefRay was not passed any parameter. It has been fixed.

Also, there is a final problem for assigning colors when drawing the rays: we must use the maximum starting height, not halfHeight.


raytracing/tests/testsImagingPath.py
```diff
@@ -189,8 +189,6 @@ def testChiefRayYIsNone(self):
         path.append(System2f(10, 10))
         path.append(Aperture(diameter=20))
         chiefRay = path.chiefRay()
-        self.assertAlmostEqual(chiefRay.y, 20, 2)
```
dccote, line 192: Humm.... Interesting test the first time around.

PR #273, opened 2020-06-15
Matrix product has the right indices

### LINKED ISSUE #272: Matrix multiplication returns a matrix with n1 = n2 = 1
I don't really know what we should do: in some cases, both matrices have the same n1 and n2, then the result should have n1 and n2. But sometimes they differ. I saw it happen in `MatrixGroup` when we append and build a transfer matrix: we create an identity matrix, but it has n1 = n2 = 1, and when we multiply with an element of the group (which can have n1 != n2 and both != 1), the product returns a matrix with the right ABCD components, back and front vertex, but don't have the right indices of refraction.

Since `MatrixGroup` already checks if there is a mismatch between indices, we could apply the indices of the right-hand side (or left-hand side, or the one that is not 1, I don't know) of the product. But at the same time we assume something, hence `__mul__` becomes less general. @dccote what do you think we should do?

Here is an example:
````python
elements = []
elements.append(DielectricInterface(n1=1, n2=n1, R=R1, diameter=diameter))
elements.append(Space(d=tc1, n=n1))
elements.append(DielectricInterface(n1=n1, n2=n2, R=R2, diameter=diameter))
elements.append(Space(d=tc2, n=n2))
elements.append(DielectricInterface(n1=n2, n2=1, R=R3, diameter=diameter))
group = MatrixGroup(elements) # This fails because of the transfer matrix in append
````
`append` code:
````python
def append(self, matrix):
    transferMatrix = Matrix(A=1, B=0, C=0, D=1)
        distance = upTo
        for element in self.elements:
            if element.L <= distance:
                transferMatrix = element * transferMatrix # This is the problem, * keeps default indices value
                distance -= element.L
            else:
                transferMatrix = element.transferMatrix(upTo=distance) * transferMatrix
                break

        return transferMatrix
````
`mul_matrix` code
````python
def mul_matrix(self, rightSideMatrix):
    a = self.A * rightSideMatrix.A + self.B * rightSideMatrix.C
    b = self.A * rightSideMatrix.B + self.B * rightSideMatrix.D
    c = self.C * rightSideMatrix.A + self.D * rightSideMatrix.C
    d = self.C * rightSideMatrix.B + self.D * rightSideMatrix.D
    L = self.L + rightSideMatrix.L
    fv = rightSideMatrix.frontVertex
    if fv is None and self.frontVertex is not None:
        fv = rightSideMatrix.L + self.frontVertex

    if self.backVertex is not None:
        bv = rightSideMatrix.L + self.backVertex
    else:
        bv = rightSideMatrix.backVertex

    return Matrix(a, b, c, d, frontVertex=fv, backVertex=bv, physicalLength=L) # We should not keep n1 = n2 = 1

````

### PR DESCRIPTION
**Pull request by GabGabG**
I modified `mul_matrix` to consider the indices of refraction of the two matrices for the product. We were returning a matrix with the right ABCD and vertices, but the indices were kept at 1 (default value). Now, we have (when both matrices are not identity matrices):
- Front index is the front index of the left hand side of the product (i.e. `self`)
- Back index is the back index of the right hand side of the product

If RHS is identity:
- Front and back indices are the same as LHS

If LHS is identity:
- Front and back indices are the same as RHS

I made it this way, because identity matrices (such as apertures or "the first transfer matrix" in `transferMatrix`) are not supposed to affect the indices of refraction. They are just there to add a finite diameter or to be a neutral element.

We don't check for any mismatch because I think we need to keep this method as general as possible.

Fixes #272

### DIFF
--- a/raytracing/matrix.py
+++ b/raytracing/matrix.py
@@ -137,6 +137,10 @@ class Matrix(object):
         super(Matrix, self).__init__()
 
     @property
+    def isIdentity(self):
+        return self.A == 1 and self.D == 1 and self.B == 0 and self.C == 0
+
+    @property
     def determinant(self):
         """The determinant of the ABCD matrix is always frontIndex/backIndex,
         which is often 1.0
@@ -269,8 +273,17 @@ class Matrix(object):
         else:
             bv = rightSideMatrix.backVertex
 
-        return Matrix(a, b, c, d, frontVertex=fv, backVertex=bv, physicalLength=L,
-                      frontIndex=rightSideMatrix.frontIndex, backIndex=self.backIndex)
+        if self.isIdentity:  # If LHS is identity, take the other's indices
+            fIndex = rightSideMatrix.frontIndex
+            bIndex = rightSideMatrix.backIndex
+        elif rightSideMatrix.isIdentity:  # If RHS is identity, take other's indices
+            fIndex = self.frontIndex
+            bIndex = self.backIndex
+        else:  # Else, take the "first one" front index and the "last one" back index (physical first and last)
+            fIndex = rightSideMatrix.frontIndex
+            bIndex = self.backIndex
+
+        return Matrix(a, b, c, d, frontVertex=fv, backVertex=bv, physicalLength=L, frontIndex=fIndex, backIndex=bIndex)
 
     def mul_ray(self, rightSideRay):
         r"""This function does the multiplication of a ray by a matrix.
--- a/raytracing/matrix.py
+++ b/raytracing/matrix.py
@@ -269,8 +269,8 @@ class Matrix(object):
         else:
             bv = rightSideMatrix.backVertex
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by GabGabG** [comment]
See updated first comment / description

File: raytracing/matrix.py [hunk 1]
```diff
@@ -136,6 +136,10 @@ def __init__(
         self.isFlipped = False
         super(Matrix, self).__init__()
 
+    @property
+    def isIdentity(self):
+        return self.A == 1 and self.D == 1 and self.B == 0 and self.C == 0
```
dccote, line 141: Understand the reasoning: if it is identify, then we should do that.  But I think the source of all that is that the index is really "undetermined" more than anything.  I wonder what would happen if we set index to None by default?

GabGabG: It is something possible, but it requires a lot of changes, everywhere: in the doc, we assume a lot of times that n = 1 by default, so we have to change it. `Lens`, `CurvedMirror` assume that n = 1 by default.

PR #280, opened 2020-06-18
Fix display of demos

### LINKED ISSUE #279: demo #3-13-16-18-19 not working with 1.2.10
I installed raytracing with pip.
i called "python -m raytracing"
demo 1 and 2 worked, but not 3, so others didnt display.
i called every demo 1 by 1 to see which one is working.
It looks like some demos are not displaying correclty (i think).

JLBegin:
fix #279

- Block the use of 'onlyPrincipalAndAxialRays' if ImagingPath.principalRay() is None
- Fixed ThickLens graphic
- Fixed many displayRange and plt scaling issues

#### update
- Fixed graphic for subclasses of specialty lenses
- Added display method in Matrix() which calls Graphic(self) to display any Matrix or MatrixGroup subclasses correctly.

### DIFF
--- a/raytracing/__main__.py
+++ b/raytracing/__main__.py
@@ -393,7 +393,7 @@ if 18 in examples:
     path.append(Space(d=180))
     path.append(olympus.LUMPlanFL40X())
     path.append(Space(d=10))
-    path.display(inputBeam=GaussianBeam(w=0.001), comments="""
+    path.display(beams=[GaussianBeam(w=0.001)], comments="""
     path = LaserPath()
     path.label = "Demo #18: Laser beam and vendor lenses"
     path.append(Space(d=50))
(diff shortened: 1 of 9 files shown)


dccote:
The `AchromaticDoublets` do not have their `display()` function back.
`python -m raytracing -e 16`
AttributeError: 'AC254_050_A' object has no attribute 'display'

JLBegin:
I added the display function in Matrix(). It calls Graphic(self).display() which will display the corresponding graphic based on the class type.

JLBegin:
There's still a bug with demo 19. It uses LaserCavity but laserModes returns an empty list.

dccote:
> There's still a bug with demo 19. It uses LaserCavity but laserModes return an empty list.

I will look into that

dccote:
The radius of curvature of the curved mirror had the wrong sign and was changed in 1.2.8, but the example was not modified.

PR #293, opened 2020-06-25
Axicon  is fully tested and fixed some problems

### LINKED ISSUE #291: Axicon does not change the values of backIndex and frontIndex
The `Axicon` class does not change the default values of `backIndex` and `frontIndex` with the input value `n`.

### LINKED ISSUE #292: axicon does not override mul_matrix
`Axicon` should override `mul_matrix`, but instead there is a method called `mul_mat` which *does* what `mul_matrix` should do. That way, `Axicon.mul_mat(...)` works like it should (raises an exception), but `Axicon.__mul__(...)` does not raise any exception when the argument is a `Matrix`.

GabGabG:
There were not tests for `Axicon`, so I decided to do some. I also found some issues that I fixed.

Fixes #291 and Fixes #292

### DIFF
--- a/raytracing/axicon.py
+++ b/raytracing/axicon.py
@@ -4,9 +4,11 @@ import matplotlib.pyplot as plt
 
 class Axicon(Matrix):
     """
-    This class is an advanced module that describes an axicon lens, not part of the basic formalism.
-    Using this class an axicon conical lens can be presented.
+    This class is an advanced module that describes an axicon lens, 
+    not part of the basic formalism. Using this class an axicon 
+    conical lens can be presented.
     Axicon lenses are used to obtain a line focus instead of a point.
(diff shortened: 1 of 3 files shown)


dccote:
A few mistakes: 
1. the index around should be air, because that is how the deviation angle is calculated.
2. All angles are in radians, I added a `degrees` constant equals to 3.14/180. 2*degrees gives 2 degrees in radians.
3. The ray from below (negative angle) must go up and vice versa.
4. Geometric considerations are used to estimate `focalLineLength` for the tests.

dccote:
> I understand the changes. Axicon is really an interesting object! I was not sure is the gaussian beam multiplication was still relevant with the axicon, but since it was not overriden, I supposed it was.

I had forgotten that because I added GaussianBeams after I did the Axicon.

PR #297, opened 2020-06-26
Creation of `utils.py` for recovering performance

dccote:
The numpy function `isclose` is actualy surprinsingly slow.  I was able to create a smalleer function that is 10x faster.  Since we are using it in `Matrix.__init__`, it is critical to be fast and as it stood, it slowed down the plotting of LaserPath and others significantly.  I created 4 functions:
`isAlmostZero()`, `isAlmostOne()`, `areAlmostEqual()` and `areNotEqual()`

These functions are now in a new file called `utils.py`, with a few other functions.

I was not able to create a test to measure the performance before and after. However, I did use:
`python -m cProfile -s time -m raytracing -e 19 | more`

to show that `__init__` went from 2 seconds to 0.144 seconds:
```
    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    1081    1.746    0.002    1.791    0.002 {method 'start_event_loop' of '_macosx.FigureCanvas' objects}
    1081    0.470    0.000    0.537    0.000 {method 'show' of '_macosx.FigureManager' objects}
    81434    0.137    0.000    0.311    0.000 matrix.py:199(mul_matrix)
    82253    0.121    0.000    0.144    0.000 matrix.py:102(__init__)
```

### DIFF
--- a/raytracing/gaussianbeam.py
+++ b/raytracing/gaussianbeam.py
@@ -45,7 +45,7 @@ class GaussianBeam(object):
             raise ValueError("Please specify 'q' or 'w'.")
 
         if q is not None and w is not None:
-            if areNotEqualRelative(self.q, q, relTol):
+            if areRelativelyNotEqual(self.q, q, relTol):
                 msg = f"Mismatch between the given q '{q}' and the computed q '{self.q}' ({relTol * 100}% tolerance)."
                 raise ValueError(msg)
 
--- a/raytracing/gaussianbeam.py
+++ b/raytracing/gaussianbeam.py
@@ -1,6 +1,6 @@
 import math
 import cmath
-
+from .utils import *
 
 class GaussianBeam(object):
     """A gaussian laser beam using the ABCD formalism for propagation of complex radius of curvature q.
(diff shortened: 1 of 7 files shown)


raytracing/gaussianbeam.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, q: complex = None, w: float = None, R: float = float("+Inf"),
             raise ValueError("Please specify 'q' or 'w'.")
 
         if q is not None and w is not None:
-            if not cmath.isclose(a=self.q, b=q, rel_tol=relTol):
+            if areNotEqual(self.q.real, q.real) or areNotEqual(self.q.imag, q.imag):
```
GabGabG: Is it possible to have a *relative* difference instead of an absolute one? If real values or imag values are huge, an absolute difference of, let's say epsilon = 1e- 3 doesn't make sense.

dccote: Yes but we will need to be careful : relative when the value is near 0 cannot be done easily.
We should create `areAbsolutelyNotEqual()` and `areRelativelyNotEqual()` for instance.

GabGabG: I created relative methods, but I'm not sure if they are robust enough when close to 0.

PR #298, opened 2020-06-26
ImagingPath's ray fan variables should not be used. They are deprecated.

### LINKED ISSUE #266: `ImagingPath` should use Rays() instead of object and fanAngle etc...
We have the right mechanism to create multiple rays of appropriate distribution.  We should use an instance of Rays() instead of separate variables in `ImagingPath`.

GabGabG:
**This is still a work on progress. I want to get your feedback on this before I dive too much in it**

Since we have classes such as `UniformRays`, `RandomRays`, etc, using `Ray` static methods `fan` and `fanGroup` is not really recommended anymore. Hence, we should not use multiple variables in `ImagingPath` (`fanAngle`, `fanNumber`, `rayNumber`).

I implemented properties getter / setter for those 3 attributes. With that, we don't change the API and it allows us to send warnings about the deprecation of them. It adds quite some code lines, but eventually in a future version we could simply remove the variables and remove said code lines. Current users are now warned that they should change their code because we will eventually remove `fanAngle`, `fanNumber`, `rayNumber`. I searched a bit on internet to know if there was a simpler way to add deprecation warnings to attributes, but I didn't find anything simpler.

I also created a new property, `rays`, which is a `UniforRays` instance. It is initialized in the constructor with default values and it can be updated with `updateRays` method. This method takes `yMax`, `yMin`, `thetaMax`, `thetaMin`, `M`, `N` to create new rays.

I did not test it before creating this <s>pull</s> review request. I also did not change anything related to demos that involves `fanAngle`, `fanNumber`, `rayNumber`. I want to know what you think before I modify too many lines.

Might fix #266

### DIFF
--- a/raytracing/figure.py
+++ b/raytracing/figure.py
@@ -559,12 +559,12 @@ class Figure:
         else:
             halfAngle = self.path.fanAngle / 2.0
             halfHeight = self.path.objectHeight / 2.0
-            rayGroup = Ray.fanGroup(
+            rayGroup = UniformRays(
                 yMin=-halfHeight,
                 yMax=halfHeight,
                 M=self.path.rayNumber,
-                radianMin=-halfAngle,
-                radianMax=halfAngle,
+                thetaMin=-halfAngle,
+                thetaMax=halfAngle,
                 N=self.path.fanNumber)
             linewidth = 0.5
 
--- a/raytracing/imagingpath.py
+++ b/raytracing/imagingpath.py
@@ -82,9 +82,11 @@ class ImagingPath(MatrixGroup):
 
         self._objectHeight = 10.0  # object height (full).
         self.objectPosition = 0.0  # always at z=0 for now.
-        self.fanAngle = 0.1  # full fan angle for rays
-        self.fanNumber = 9  # number of rays in fan
-        self.rayNumber = 3  # number of points on object
+        self._fanAngle = 0.1  # full fan angle for rays
+        self._fanNumber = 9  # number of rays in fan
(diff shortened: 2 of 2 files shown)


GabGabG:
The code might not be as optimized as it can be. It is only a draft to show what's on my mind about issue #266

dccote:
I am thinking a better option is to pass `Rays` to `display()`.  The `objectHeight` and others are never really used until the whole thing is displayed. It caused confusion to other users: `fieldOfView()`, `objectHeight`, which one is used?

I like the idea of completely removing this aspect that is *totally* display-related.

We could have:
```
display(self, rays=None)
   # if rays is None, use principalAndAxialRays with objectHeight = field of view

displayWithObject(self, height, fanAngle, fanNumber, rayNumber):
    # create Rays() and call self.display(rays)

```
or something like that to move the object and ray tracing "out" of the ImagingPath. Do you think it makes sense?

GabGabG:
I like the idea to separate the display stuff from the intern computing stuff. I feel like it would be easier to maintain and easier to work with.

dccote:
> I like the idea to separate the display stuff from the intern computing stuff. I feel like it would be easier to maintain and easier to work with.

Will you have time to do it? If not, I could give it a try.

GabGabG:
> Will you have time to do it? If not, I could give it a try.

Isn't it what `figure.py` is supposed to do already? I thought the process of separating the computations from the display was already in motion. If not, I can work on that, no problem.

dccote:
I made a mistake and started another branch.  I think this branch will become dated and redundant.

PR #299, opened 2020-06-28
Syntax error : use of is instead of ==, fixed

dccote:
In `figure.py`, line 115 and 116 were written with "is", which is ok with 3.8 but not 3.6.  I think it is better to maintain 3.6 compatibility.
Incorrect justification: see below


JLBegin:
The "is" keyword works for all python versions (and I code with 3.6). And I dont think its usage has changed. But "is" checks two object types while "==" checks two values. So in this case there was an error in the code since I want to check the value with "==". I would keep the changes of this PR but change the title.

PR #300, opened 2020-06-30
Lagrange and optical invariant improvements

dccote:
This is a work in progress. Some things to add up.

Several aspects related to computing and using the Lagrange invariant have been modified.

First, the general name when used with any two rays is the opticalInvariant, not the Lagrange invariant.
Second, the Lagrange invariant is, by definition, the optical invariant for the principal and axial rays.
Therefore, a new function has been created called `opticalInvariant()` for `Matrix`, and `ImagingPath` is the only one with `lagrangeInvariant()`, because it si the only one that has principal and axial rays. `lagrangeInvariant()` does not take any arguments, because it must be the principal and axial rays.

Then, a `reportEfficiency()` function has. been created to compute the overall efficiency of the system with respect to absolute and relative values.  *This is not fully ready yet*.

To help in understanding the various implications of the optical invariant and the Lagrange invariant, a new method called the `maximumInvariant()` has been created: it will calculate the maximum value of an optical invariant that can be supported by the element.  Currently, only `Objective` have it defined.
Objectives also really need the field number and the expected magnification, which have been added to the `__init__` function, tests have been modified.

A bit by distraction, a fix for the `doctest` package has been included: `numpy.matrix` was in conflict with `raytracing.matrix`.
@GabGabG Also, a fix for `subprocess.call()` has been included (`subprocess.run()` instead) because it crashed on macOS.

The main file where some "experiments" are conducted is `testsLagrangeInvariantIdeas.py`.

Questions that remained to be answered: 
1. The maximumInvariant in an objective does not appear to be ... the maximum (!).  I created a simple objective+ tube lens setup, and the invariant does not behave like I expected.
2. Not sure how to compute efficiency.
3. Some nomenclature must be improved (height vs diameter, size etc...)
4. Everything was dumped into reportEfficiency, but graphing functions should be elsewhere.  In addition, the key computations should be made into separate functions for `Matrix` so users can use them.


raytracing/tests/testsLagrangeInvariantIdeas.py
```diff
@@ -0,0 +1,81 @@
...
+    @envtest.skip
+    def testLagrange(self):
+        path = ImagingPath()
+        path.objectHeight = 50
+        path.append(System4f(f1=30, diameter1=60, f2=40, diameter2=100))
+        path.append(Aperture(diameter=10, label='Camera'))
+        # path.display()
+        path.reportEfficiency(nRays=10000)
+
+    def testObjective(self):
```
GabGabG, line 67: Since this test displays some graphs and uses `pyplot.show()`, it displays also other plots that were not showed. For instance, in `testsImagingPath.py`, the test `testSave` that tests if a figure is really saved, creates the subplots but doesn't show them. But with this test, `path.display()` displays **every** undisplayed plot. I implemented `tearDown` in `envtest.RaytracingTestCase` to clear plots after every test. See the commits I have done.

Issue #238, opened 2020-06-06
Example : Laser Scanning Microscopy


Valeriouus:
Here is my code for now. 

```
import envexamples
from raytracing import *

# A polygonal mirror with 36 facets has an angle of 10 degrees, 0.1750 rad, between the facets.

class LUMPlanFL40X(Objective):

    def __init__(self):
        super(LUMPlanFL40X, self).__init__(f=180/40,
                                         NA=0.8,
                                         focusToFocusLength=40,
                                         backAperture=7,
                                         workingDistance=2,
                                         label='LUMPlanFL40X Objective',
                                         url="https://www.edmundoptics.com/p/olympus-lumplfln-40xw-objective/3901/")


illumination = ImagingPath()
# Let's say that the laser has a diameter of 500 nm.
illumination.objectHeight = 0.0005
illumination.rayNumber = 3
illumination.fanNumber = 3
illumination.fanAngle = 0
illumination.append(System4f(f1=40, f2=75, diameter1=24.5, diameter2=24.5))
illumination.append(System4f(f1=100, f2=100, diameter1=24.5, diameter2=24.5))
illumination.append(Space(d=180/40))
illumination.append(LUMPlanFL40X())
illumination.append(Space(d=180/40))
illumination.display()

```

- I need to change the angle of which the rays are propagating at the object plane
- I would like to make a list of the image position in y for each object angle.

shadimasoumi:
The text is done!

Valeriouus:
The text is done in the article. The code is done in the article and in the _examples_ folder. Please tell me your comments! :)

Valeriouus:
I deleted the previous graph and kept Daniel's changes. Also, the text has been changed in the article. Plz review!

PR #303, opened 2020-06-30
Laser scanning microscope

Valeriouus:
Text is done in the article. The code is in the article and in the _examples_ folder. Please take a look and tell us you comments!
#238


dccote:
I think it would be more useful to print the results to screen rather than try to show the graph. We could simply have something like:
```
...

ray = Ray(0, 10*np.pi/180)
rayAtFocus = illumination.traceThrough(ray)
print("Scan angle of 10 degrees gives a displacement of {0:.1f} µm".format(rayAtFocus.y*1000))

axialRay = illumination.axialRay()
maxRayAtFocus = illumination.traceThrough(axialRay)
print("Maximum scan is {0:.1f} degrees for a displacement of {0:.1f} µm".format(axialRay*180/np.pi, maxRayAtFocus.y*1000))

```

dccote:
I added this in the file:
```
thetas = []
positions = []

path = illuminationPath()
scanRays = UniformRays(yMax=0, thetaMax=10*3.14/180, M=1, N=10)
outputRays = path.traceManyThrough(scanRays)
for i in range(len(outputRays)):
    thetas.append(scanRays[i].theta*180/np.pi)
    positions.append(outputRays[i].y*1000)

plt.plot(thetas, positions)
plt.ylabel('Scanning position of the focal spot (µm)')
plt.xlabel('Scan angle (degrees)')
plt.show()
```

PR #313, opened 2020-07-13
Change max Invariant symbol to make it correspond to the article

Valeriouus:
see reportEfficiency() in imagingpath.py. Change Iap for Zhe.


dccote:
I prefer if we use a comment to refer to the symbol from the article.  I don't like the use of weird variable names in the code.

Valeriouus:
Done. Added a comment when Iap is defined.

PR #320, opened 2020-07-16
Improve fontsizes

JLBegin:
- Setting all fontsizes so they are not set automatically based on a user's screen (I'm not a 100% sure of that)
- Linked all fontsizes to a figure.fontScale that can be adjusted using the design method. Default to 1.

This is a draft, please comment which fonts to change, or if you have other ideas on how to handle the fonts. We can always adjust the overall fontscale but I would like to get a proper ratio of all fonts. 

So let's look at the following code. 
```
path = ImagingPath()
path.label = "Demo #5: Simple microscope system"
path.append(Space(d=4))
path.append(Lens(f=4, diameter=0.8, label='Obj'))
path.append(Space(d=4 + 18))
path.append(Lens(f=18, diameter=5.0, label='Tube Lens'))
path.append(Space(d=18))
path.display()
```


shadimasoumi:
Thanks!
The upper pic is obtained with previous setting and the lower one is done with the new setting!
(I put them in the article draft to see how they look like)
I do not see any obvious difference between these two pictures.
So if I did right, the font size of all (axis numbers, labels and ...) should be larger!
Screen Shot 2020-07-16 at 5 12 58 PM

JLBegin:
> Thanks!
> The upper pic is obtained with previous setting and the lower one is done with the new setting!
> (I put them in the article draft to see how they look like)
> I do not see any obvious difference between these two pictures.
> So if I did right, the font size of all (axis numbers, labels and ...) should be larger!

(Ok this also shows another issue: rays.display is a standalone and does not use our Figure manager.)
Anyways for now I fixed some font sizes for the rays.display. It should be the same accross different screens and bigger. @shadimasoumi  can you test it again? Also if you could check the fontsizes for the display of an ImagingPath that would be great. Thank you

shadimasoumi:
> > Thanks!
> > The upper pic is obtained with previous setting and the lower one is done with the new setting!
> > (I put them in the article draft to see how they look like)
> > I do not see any obvious difference between these two pictures.
> > So if I did right, the font size of all (axis numbers, labels and ...) should be larger!
> 
> (Ok this also shows another issue: rays.display is a standalone and does not use our Figure manager.)
> Anyways for now I fixed some font sizes for the rays.display. It should be the same accross different screens and bigger. @shadimasoumi can you test it again? Also if you could check the fontsizes for the display of an ImagingPath that would be great. Thank you

@JLBegin  It does work for ImagingPath and the fontsizes have been enlarged well! but for rays.display it is the same as it was before!!! no changes!

JLBegin:
ok @shadimasoumi could you test again and send a picture if possible?

shadimasoumi:
Of course!

imagingpath

raypath

shadimasoumi:
The upper figures are tested in the branch you are working on and the following pic is tested in the master!

JLBegin:
Ok now I personally have these two displays. @shadimasoumi what do you think, and how does it compare with what you have on your end ?
image
image

shadimasoumi:
@JLBegin  I tried some other examples and It looks perfect.

PR #328, opened 2020-07-28
Improve LampRays and ObjectRays 

JLBegin:
## Changes
- LampRays is now by default a linear distribution of rays (accross 'y' and 'theta' to fill diameter and NA). 
- Added `random` arg in LampRays to get a random distribution (as before). 
- Added an edge case for H=1 when creating UniformRays (ObjectRays) so that the fan is created at y=0 and not y=-yMin
- Remove figure text note about Object Height since we are now always talking about FOV. Keep FOV and Image size text note.

### LampRays
with default argument `random=True`. We can also see that it now writes FOV size instead of object height
image

```
path = ImagingPath()
path.append(Space(d=10))
path.append(Lens(f=10, diameter=100, label="Collector"))
path.append(Space(d=10+30))
path.append(Lens(f=30, diameter=100, label="Condenser"))
path.append(Space(d=30+30))
path.append(Lens(f=30, diameter=100, label="Objective"))
path.append(Space(d=30+30))
path.append(Lens(f=30, diameter=100, label="Tube"))
path.append(Space(d=30+30))
path.append(Lens(f=30, diameter=100, label="Eyepiece"))
path.append(Space(d=30+2))
path.append(Lens(f=2, diameter=10, label="Eye Entrance"))
path.append(Space(d=2))

path.design(lampRayColors='y')
path.display(raysList=[LampRays(15, N=50, random=False),
                       ObjectRays(40, halfAngle=0, T=1, z=20, rayColors='r', color='r'),
                       ObjectRays(30, halfAngle=0, T=1, z=80, rayColors='g', color='g')], removeBlocked=False)
```

### ObjectRays(H=1)
image

```
path = ImagingPath()
path.design(lampRayColors='r')
path.append(Space(d=10))
path.append(Lens(f=5, diameter=3))
path.append(Space(d=3))
path.append(Aperture(diameter=3))
path.append(Space(d=17))
path.display(ObjectRays(5, 0.1, H=1),
             removeBlocked=False)
```

### DIFF
--- a/raytracing/figure.py
+++ b/raytracing/figure.py
@@ -116,9 +116,6 @@ class Figure:
                 warnings.warn("Infinite image size: cannot use limitObjectToFieldOfView=True.")
                 self.designParams['limitObjectToFieldOfView'] = False
 
-        if not self.designParams['limitObjectToFieldOfView']:
-            note1 = "Object height: {0:.2f}".format(self.path.objectHeight)
-
         if self.designParams['onlyPrincipalAndAxialRays']:
             (stopPosition, stopDiameter) = self.path.apertureStop()
             if stopPosition is None or self.path.principalRay() is None:
--- a/raytracing/rays.py
+++ b/raytracing/rays.py
@@ -777,6 +777,7 @@ class RandomLambertianRays(RandomRays):
         self.append(ray)
         return ray
 
+
 class ObjectRays(UniformRays):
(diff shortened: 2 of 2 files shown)


Valeriouus:
Tiguedou ça!

dccote:
I'm not sure I like that for the `Lamp`: are the measurements still ok at the various Fourier planes? I get the feeling they are not. It looks good though.

PR #341, opened 2020-08-18
RandomLambertianRays : add theta arguments

JLBegin:


JLBegin:
Good, I thought it could be that

PR #361, opened 2021-01-07
Issue#103 automatichierarchy

dccote:
What is the status on this?


PyMarc2:
It was very confusing and not really worth it. Spent way too much time on this. We have to use a kind of sphinx plugging. it works, but the result is really not that impressive. Doing a manual hierarchy would almost be better.

dccote:
So we kill the pull request and associated branch?

Issue #363, opened 2021-01-08
Objectives not displaying and new arguments missing


FrancoisCoteIng:
I wanted to display a Nikon objective using the specialityLenses class "Objective". It seems that there are 2 new arguments needed to define an objective: Field Number (which is not specified by all manufactures) and magnification (which depend on the tube lens used, so I don't see the point to ask for it if the focal is already in the definition). Even if I add those new arguments, it seems that the algorithm doesn't want to draw my objective and can't flip it using .flipOrientation().

dccote:
The guilty will be found and hanged.

dccote:
@FrancoisCoteIng We need example code to reproduce this, since the `python -m raytracing` examples work fine.

FrancoisCoteIng:
try:
    from raytracing import *
except ImportError:
    raise ImportError('Raytracing module not found: "pip install raytracing"')

import nikonObj as nikon

class Confocal:
    @staticmethod
    def detection():
        detection = ImagingPath()
        detection.label = "VMS/Confocal Base detection"
        detection.objectHeight = 0.5+0.2 # mm maximum, include diffuse spot size
        detection.fanAngle = 1.1 # NA = 1.1
        detection.fanNumber = 11
        detection.rayNumber = 3
        detection.showImages = False

        fTubeLens=200
        fScanLens=54
        f1=150
        f2=75

        obj = nikon.N25ApoMD()
        obj.flipOrientation()

        detection.append(Space(d=2))
        detection.append(obj)
        detection.append(Space(d=fTubeLens)) 
        detection.append(Lens(f=fTubeLens,diameter=50, label='Tube Lens'))
        detection.append(Space(d=fTubeLens))
        detection.append(Space(d=fScanLens))
        detection.append(Lens(f=fScanLens, diameter=30, label='Scan Lens'))
        detection.append(Space(d=fScanLens))

        return detection


if __name__ == "__main__":

    test = Confocal.detection()
    test.showImages = True
    test.objectHeight = 1
    test.rayNumber = 5
    test.fanNumber = 5

    test.display()

FrancoisCoteIng:
nikonObj.py :

from raytracing.specialtylenses import *

class LWD16X(Objective):
    """ Nikon 16x immersion objective

    .. csv-table::
        :header: Parameter, value

        "f", "200/16"
        "backAperture", "20"
        "Numerical Aperture (NA)", "1.56"
        "Working Distance (mm)", "3"

    Notes
    -----
[...]

PR #365, opened 2021-01-22
Fix flipped graphics of surface-based elements and objectives

JLBegin:
*Requires review

### Fix for flipped graphics
**Surfaces**
"renamed" all `surfaces` property occurences for `forwardSurfaces`. Redefined `surfaces` property in Matrix baseclass which returns `forwardSurfaces` unless matrix or matrixgroup `isFlipped` where it reinitiates the surfaces in a flipped order (correcting the medium index & length). 

**Objectives** (#364 and part of #363)
Simply flipping the default polygon graphic.

### Examples
**Example display for a surface-based element (AchromatDoubletLens) and an objective. On the left is the 'forward' setup, and on the right is the flipped version. While the graphics look fine I am not sure of the properties.**

flip

### DIFF
--- a/raytracing/axicon.py
+++ b/raytracing/axicon.py
@@ -132,7 +132,7 @@ class Axicon(Matrix):
         raise TypeError("Cannot use Axicon with GaussianBeam, only with Ray")
 
     @property
-    def surfaces(self):
+    def forwardSurfaces(self):
         """ A list of surfaces that represents the element for drawing purposes
         """
         minThickness = - np.tan(self.alpha) * self.displayHalfHeight()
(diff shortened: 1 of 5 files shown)


dccote:
I wonder what we should do with the name "front" and "back": the lens does have a "front" side and a "back" (from its design).  So when we flip it, should we indicate "this lens is flipped" and leave front and back as designed (so they will appear wrong), or rename everything (like done here) and put "front" and "back" as expected (like done here?).

dccote (approved):
We'll take it like that for now, but I am not sure what to do with with front/back regarding the comment i made earlier: when flipped, do we keep the names front and back as they were designed, or as flipped and shown?

PR #377, opened 2021-02-22
Named tuples for magnification, focal lengths and principal planes.

dccote:
`Magnification`, `FocalLengths`, `PrincipalPlanes`, `Stop`, `MarginalRays` and others were defined as namedtuples to make manipulation and meaning easier for users.

I wonder if there are better "NamedTuples" that could be more general, but so far this is the best I found.


JLBegin (approved):
Nice use of NamedTuples

PR #380, opened 2021-02-23
Add deprecation warnings for all object references in ImagingPath except objectHeight.

dccote:
`fanAngle`, `fanNumber` and `rayNumber` in `ImagingPath` are not necessary: we have `ObjectRays` to create default rays for an object.  The only variable needed is `objectHeight`, which was kept.

### DIFF
--- a/raytracing/imagingpath.py
+++ b/raytracing/imagingpath.py
@@ -77,6 +77,10 @@ class ImagingPath(MatrixGroup):
     def __init__(self, elements: list = None, label=""):
 
         self._objectHeight = 10.0  # object height (full).
+        self.objectPosition = 0.0  # always at z=0 for now.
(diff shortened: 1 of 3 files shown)


dccote:
Crap: it is mentionned in the illumination example on page 14 of https://doi.org/10.1117/1.NPh.8.1.010801. Therefore we cannot remove it now, we will need to add a deprecation warning instead and remove it in 1.4.x

I reverted the commit and added a deprecated warning.

JLBegin:
Added deprecation warnings to fanAngle, fanNumber, rayNumber:

"DeprecationWarning:Object references (fanAngle, fanNumber, rayNumber) will be removed from ImagingPath in future versions. Create an ObjectRays(...) instead and provide it to the display with ImagingPath.display(rays=...)"

PR #386, opened 2021-03-10
Bugfix: multiprocessing falls back to single if it fails

dccote:
Fall back to single-processing code when multiprocessing code fails.


GabGabG (approved):
Good idea.

PR #389, opened 2021-04-27
Added a simple graph to show chromatic aberrations of lenses when possible

dccote:
Simple function to show chromatic aberrations for a lens when possible, that is, when we know the material it is made of.
Only possible with `AchromaticDoublet` and `SingletLens`. Was tested with AC254_100_A(), appears essentially identical to the spec (but not perfectly as specified)

```
from raytracing import *

thorlabs.AC254_100_A().showChromaticAberrations()
```
This graph is "the testing graph":
image

and this graph is the actual graph from the module:
image

In the future, if we want to include more chromatic aberrations, dispersions, etc... we will need to move the wavelength dependence to base functions: right now, when we create the thorlabs.lens, for instance, we set the wavelength. It would be better to give a wavelength to `Ray` and use that wavelength in calculations when obtaining focal lengths etc....  This is a significant overhaul of the code and will not be done now. But when we do, we will be able to compute the chromatic aberrations of the entire optical system.


PyMarc2 (approved):
All good from me. Excited to participate in the aberration calculations in the future.

JLBegin (approved):

PR #392, opened 2021-04-28
Spherical aberrations

dccote:
Spherical aberrations are in fact, surprisingly simple to implement because they simply come from the non paraxial rays. They arise because the approximation sin theta approx theta is not valid anymore. Because it is not valid, it is not possible to get a linear expression for theta' as a function of y and theta. Another way to look at it is that the transfer matrix is not constant, it depends on the height of the ray.

So instead of computing a ray transform with the paraxial form of the matrix, we offer two options:
1. Use the paraxial approximation and the paraxial matrix.
2. Use non-paraxial rays and compute the exact refracted ray from Snell's law.
3. Get a non-paraxial matrix that depends on y.

So I split `mul_ray` into two fonctions: `mul_ray_paraxial` and `mul_ray_nonparaxial`.  Currently, a global setting in a class variable `Matrix.useNonParaxialCalculations = False` is used, but it looks a bit out of place.

We have `AsphericDielectricInterface` with a `kappa` conic parameter. It derives from `DielectricInterface` because it is a natural extension, and `DielectricInterface` is the paraxial limit of `AsphericDielectricInterface`. Since we have the complete equation for the surface z(y), we can compute the angle of incidence, the refraction and everything exactly.

We still don't move to the exact point in z on the surface because it is incompatible with the element that follows: if we need to move by Space(10), then the matrix will compute the displacement assuming that the ray is at its front vertex and move from frontVertex to backVertex, but that would be incorrect. I may try to add a small correction, but not sure exactly how to do that in a way that is clean and elegant.


dccote:
Major problem: we cannot consider spherical aberrations (non paraxial calculations) without also considering the finite thickness of elements (i.e. the fact that the distance to travel when rays are significantly off-axis is different from the paraxial rays on or close to the axis) . If we go ahead with the present partial implementtion, it would appear to be considering the aberrations, but without correcting for the varying distance as we move up the element, the numbers would be wrong., so there is no point providing a "half-baked" aspherics.

However, I think there is a way of doing it: the problem comes from the fact that an Interface really has a zero thickness, but an "non-paraxial Aspheric" does not: we must consider the thickness in order to have a real transfer matrix from one reference plane to the other.

So we will create some sort of "AsphericDielectric" (AsphericMeniscus?) that will have a R, kappa but also a finite, non-zero thickness.  It will be essentially a Space() Aspheric() Space() matrix group but the Space() elements will have a varying thickness depending on the ray.  In the paraxial limit., this will just be an interface to a dielectric.

raytracing/tests/testsAsphericInterface.py
```diff
@@ -0,0 +1,178 @@
...
+        for paraxialRay in rays:
+            paraxialOut = asphere.mul_ray_paraxial(paraxialRay)
+            nonparaxialOut = asphere.mul_ray_nonparaxial(paraxialRay)
+            self.assertAlmostEqual(paraxialOut.y, nonparaxialOut.y, 3)
+            self.assertAlmostEqual(paraxialOut.theta, nonparaxialOut.theta, 3)
+
+    @envtest.skip
+    def testOpticalPath(self):
+        Matrix.useNonParaxialCalculations = True
+        path = ImagingPath()
```
JLBegin, line 156: Maybe have `ImagingPath(useNonParaxialCalculations=None)` instead ? Just for usage. And I guess we keep the global in Matrix.

dccote: It should be in OpticalPath or something like that for the user, but deep down, it is at the matrix level that this affects the calculation. You are right that it could go in the ImagingPath/OpticalPath `__init__`.

See the above comment that I made essentially 2 minutes before you made the present comment: I am not sure anymore if it should be "a global choice", because I am not sure the results are correct.

PR #393, opened 2021-04-28
Building lenses from prescriptions and Zemax files (ZMX only)

dccote:
*(It is the wrong branch name by mistakem, this has nothing to do with chromatic-aberrations, I will not rename it).*

The prescription of a compound lens is the list of interfaces and materials everywhere along the axis. With DielectricInterfaces we can easily build a lens with the exact radii of curvatures and materials (or indexes).

After having found that Olympus has the lens prescription in public patents, I added a specifications directory in the project where we can experiment with building or validating lenses from prescriptions. There are many things that need to be determined before we make this a complete feature:
1. The format for the prescription, 
2. How we would name an eventual Group: would it be a `MatrixGroup` with different labels, or a different class? Currently, the lenses are a separate class each time because they can easily be used like `thorlabs.AC254_100_A()` without parameters.  Doing so would require an extra step: dynamically generating python files from the prescription files.

We found that although the Thorlabs catalog is in binary Zemax format (ZMF), it is still possible to obtain the ZMX files for individual lenses.  This is interesting because there is a Zemax reader in Python that works but only with the old format (https://github.com/quartiq/rayopt/blob/master/rayopt/zemax.py). Looking at it is enough for us to code the small parts we will need. (See below, I recoded the parts I needed.)

1. The page for each product is always a link of the form: https://www.thorlabs.com/thorproduct.cfm?partnumber=AC127-019-A
2. The corresponding link is https://www.thorlabs.com/_sd.cfm?fileName=20517-S03.zmx&partNumber=AC127-019-A

So we could download them by hand, and figure out the strategy for the file name and build the URL. *Update*: in fact, this is almost trivial: it is always the same filename for a group of lenses with a different part number.  I tried several options to request "all lenses" with wildcards but that did not work.  Does not matter! Looking good.

A separate class called `ZMXReader` can read a simple file (was only tested with AC254-100-A, my favourite lens and a random Edmund Lens). We read the curvature, the spacing, the material, we can create a MatrixGroup and write the prescription to the screen.


raytracing/tests/testsZemax.py
```diff
@@ -0,0 +1,85 @@
...
+        surface = self.zmx.surfaceInfo(index=1)
+        self.assertTrue( surface.number == 1)
+        self.assertAlmostEqual( surface.R, 62.75, 1)
+        self.assertAlmostEqual( surface.spacing, 4.0) 
+
+    def testSurfaces(self):
+        self.assertIsNotNone(self.zmx)
+        self.assertTrue(len(self.zmx.surfaces()) == 5)
+
+    def testMatrixGroup(self):
```
JLBegin, line 39: This testMatrixGroup is ignored because it is overwritten by the following test with the same name.

dccote: Fixed and test improved.

dccote:
> this is my first review!! (I ran the tests and everything was good)

Did you try to check for coverage?

PR #397, opened 2021-04-30
Axial ray line style: dashed line

dccote:
Is this ready? Please complete or delete.


dccote:
It looks weird to only have the axial ray as a dashed line.  I think the principal ray should also be a dashed line.

JLBegin:
I think this should be discarded unless there's a case for using dashed lines (or a linked issue).

dccote:
Closed because we're not sure this is needed.

PR #399, opened 2021-05-03
Cleaning up tests

dccote:
We had a few situations where tests were not ran before a release because they were getting annoying (they blocked before of graphs), and had buggy releases. We always need to run tests all the time, abnd for that we need to 
1. Tests used to call `plt.show()` and stop the flow.  We must always patch `plt.show()` to avoid stopping the tests. If we want to see the plot, then we can remove the @patch for the duration of our tests.
2. I defined `envtest.perfomanceTests` to decide whteher or not we test huge files and other long tests that are not critical.  The variable can be set by en environement variable `RAYTRACING_PERF_TESTS`. On macOS/Linux, `RAYTRACING_PERF_TESTS=True; export RAYTRACING_PERF_TESTS` will do the trick.
3. We should probably have `RAYTRACING_PLOT_TESTS` also, but not sure how we would use it. 
4. One way of doing these `plot` tests would be to have a human "accept" a given plot as "the right plot", save it, and then we could compare the bitmaps.

There remains errors in `testsDocTestModule.py`, but for some reason they are not accounted for at the end. Probably because the file does not create a testcase.


dccote:
@ElaheParham Take a look at the description, can you fix testsDocTestModule.py ?

ElaheParham:
I was thinking about the plot test for the whole module!
because some figures are plotted based on random distribution and they are not the same every time we run the test.

ElaheParham:
Also, some of the failed tests use random distribution as well (they print different values every time)

dccote:
> Also, some of the failed tests use random distribution as well (they print different values every time)

We should then provide a fixed seed to the random algorithm to make sure we always get the same values.

ElaheParham:
Does the seed that I added to RandomUniformRays is the one that you had in mind? or do you want it to be a variable that can be controlled by the user?
The problem with the documentation can be solved by adding np.random.seed(0) every where that we have used np.random

raytracing/rays.py
```diff
@@ -712,7 +712,9 @@ def randomRay(self) -> Ray:
         if len(self._rays) == self.maxCount:
             raise AttributeError("Cannot generate more random rays, maximum count achieved")
 
+        np.random.seed(0)
```
dccote, line 715: I don't think this is the way we want to do it. We want to seed the test properly, but not the code itself. So we need to seed beefore testing.  I looked at the code, and I added it to `envtest.RaytracingTestCase` so it will *always* be seeded with 0 for every single test, so we will always get the same sequence of random numbers **when testing**:

```
class RaytracingTestCase(unittest.TestCase):
    tempDir = os.path.join(tempfile.gettempdir(), "tempDir")
    def __init__(self, tests=()):
        super(RaytracingTestCase, self).__init__(tests)

    def setUp(self):
        np.random.seed(0)
```

dccote: I committed and uploaded, and I removed your `seed()` calls.

dccote: I removed your seed calls, because they would simply have forced the *same* rays to come out every single time.

dccote: Click "Resolve conversation" to confirm you have read and understood this, then we could merge?

ElaheParham (changes requested):
I have failed examples in matrix.py lines 730, 738, 788, 791 (all with random numbers!) Isn't it supposed to be the same?

dccote:
> I have failed examples in matrix.py lines 730, 738, 788, 791 (all with random numbers!) Isn't it supposed to be the same?

Did you try  checking that the random numbers are really always the same sequence?
I forgot to write a test to confirm that... :-(

dccote:
> I have failed examples in matrix.py lines 730, 738, 788, 791 (all with random numbers!) Isn't it supposed to be the same?

I get it: testDocModule.py does not use testCases...! So the seed is never called.

dccote:
Ok, I really want to merge this, I tried working on doctest for some time, but it is more complicated than I expected. Let's create a branch specifically for the doctest issue.

Issue #403, opened 2021-05-05
imageSize does not behave as expected for beginners versus advanced users


dccote:
The function `imageSize` of `ImagingPath` returns the size of the image of the field of view: it is the accepted definition in a real optical system, because the limit really is the field of view, not the object. However, in a very simple system (say a single lens with an infinite diameter), the field of view is infinite, because you need at least two finite diameters to get a finite field of view (i.e. a field stop). The following code:
```
from raytracing import *
path = ImagingPath()
path.objectHeight = 10
path.append(Space(d=100))
path.append(Lens(f=50))
print("Image size is: {0}".format(path.imageSize()))
```

will print:
`Image size is: inf`

because the FOV in this example is infinite. This unfortunately, while being correct and in line with the documentation and the definition, appears really odd to anyone who is not an expert.

After discussion, we believe the appropriate behaviour should be:

1. `imageSize` will accept an argument `objectHeight=FOV` or `useObject=True`. In this case, it will use  objectHeight or objectSize.
2. If the FOV is infinite, the module will print a warning to tell the user to use `useObject=True` to get the expected answer.
3. We are not sure if `maxImageSize` would be an approriate function to add (it would always use the FOV). We think this changes the semantic (if we add maxImageSize, what becomes the meaning of `imageSize` then?)

We need to decide:
1. What is the best name for the argument?
2. What is the best behaviour for the function? Do we all agree?
3. Is there a need for maxImageSize or is this adding confusion?

ElaheParham:
So I think the argument Name can be useObjectHeight which is optional and can be True or False, so the algorithm can be like:

If no input parameter is defined or it is False : 
        Then use FOV (Here we will have two cases)
                If FOV is infinite : 
                        Then print +Inf and a warning (this is because FOV is Inf and if you mean to use object height pay
                         attention to the input argument!)

if useObjectHeight is True then just use object height for calculation.


*********

It can be noted in the documentation that having an infinite FOV is not something that happens in
real experiments but we have considered this case for educational purposes.

dccote:
The branch `imageSizeScrewup` implements this.

PR #405, opened 2021-05-05
Image size behaviour change

dccote:
As discussed in #403, `imageSize` does not necessarily behave as expected for beginners.  

I added a default argument `useObject=False` to force the use of `self.objectHeight` instead of the field of view.  I print a warning to tell the user what to do when the FOV is infinite.


ElaheParham:
why did you use an invalid label?

ElaheParham (approved):

dccote:
> why did you use an invalid label?

I am never sure what to use, but "invalid" is "This doesn't seem right" (when you hover with your mouse)... It's not really a bug.

PR #407, opened 2021-05-05
Examples+main improvements

dccote:
`python -m raytracing` has always bugged me: it is hard to maintain and the code is hidden in `__main__`. Adding new code requires quite a bit of work.

I took the code out of `__main__` and created a submodule `examples`.  The code can actually be imported easily with:

```
import examples

examples.ex01.exemplecode()
```

examples will load all the files starting with `ex` in the directory. If people use `python -m raytracing`, then the actual source code will be displayed on the same figure in place of comments, which I think looks better than before:

image

It also prints (with syntax coloring) the code to the console.
image

You can get a list of all available examples with a description with `python -m raytracing -l`:

```
All example code on your machine is found at: /Users/dccote/GitHub/RayTracing/raytracing/examples
 1. ex01.py A single lens f = 50 mm, infinite diameter
 2. ex02.py Two lenses, infinite diameters
 3. ex03.py Finite-diameter lens
 4. ex04.py Aperture behind lens
 5. ex05.py Simple microscope system
[...]
```
@PyMarc2 @JLBegin What do you guys think? You may need to install pygments.


dccote:
I am waiting for you to release 1.3.5 :-)

examples/twoLenses.py
```diff
@@ -0,0 +1,26 @@
...
+	path.label = TITLE
+	path.append(Space(d=5))
+	path.append(Lens(f=5))
+	path.append(Space(d=20))
+	path.append(Lens(f=5))
+	path.append(Space(d=10))
+	path.display(comments=DESCRIPTION)
+
+if __name__ == "__main__":
+	exempleCode()
```
PyMarc2: pep8: end of file requires an empty line

raytracing/__main__.py
```diff
@@ -9,28 +9,50 @@
...
 examples = args['examples']
 printClasses = args['classes']
+listExamples = args['list']
 
 if printClasses:
     printClassHierarchy(Rays)
     printClassHierarchy(Matrix)
     exit(0)
-
+elif listExamples:
```
PyMarc2, line 45: Why `elif` ? there doesn't seem to be a constraint that prohibits the user of doing `-r -c`. If that is the case, we should see a `MutuallyExclusiveGroup` as the argparse argument parser class

raytracing/__main__.py
```diff
@@ -9,28 +9,50 @@
...
+    for i, name in enumerate(exampleScriptNames):
+        print("{0:2d}. {1}".format(i+1, name))
+    exit(0)
 if examples == 'all':
     examples = range(1, 30)
 else:
     examples = [int(y) for y in examples.split(',')]
 
+if 0 in examples:
```
PyMarc2: would it be `if examples:`
and removing all the other examples in the __main__ ?

raytracing/__main__.py
```diff
@@ -9,414 +9,66 @@
...
+              'figure.figsize': (10, 7),
+             'axes.labelsize': 'x-large',
+             'axes.titlesize':'x-large',
+             'xtick.labelsize':'x-large',
+             'ytick.labelsize':'x-large',
+             'font.family':'helvetica'}
+    plt.rcParams.update(params)
+
+    print("Running example code : {0}".format(showExamples))
+    for i in showExamples:
```
PyMarc2, line 67: perfect, last comment was on outdated commit. That is approved.

raytracing/examples/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+allFiles    = os.listdir(topDir)
+allFiles.sort()
+
+all = []
+for file in allFiles:
+    matchObj = re.match(r'^(ex\d+)\.py$', file)
+    if matchObj:
+        name = matchObj.group(1)
+        module = importlib.import_module(".{0}".format(name),package="raytracing.examples")
+        with open("{0}/{1}".format(topDir, file)) as f:
```
PyMarc2, line 30: clearer as f strings ? `with open(f"{topDir}/{file}") as f:`

raytracing/examples/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+            srcCode = f.readlines()
+        # The last three lines are always the main() call
+        srcCode = srcCode[:-3]
+        srcCode = ''.join(srcCode)
```
PyMarc2, line 34: `srcCode = ''.join(srcCode)` What for ?

PyMarc2 (approved):
I approve of these changes, even though the _init_ might be hard to read for some, it is worth it for the clarity of the _main_. 
However, I still prefer the way it looked with the HTML style with my old branch. The main is better managed, but the look and feel are worse, in my opinion. The title was coloured and pretty and everything seemed more "together".

I inserted a few comments that would need attention before merging

raytracing/__main__.py
dccote: It makes more sense to have only one. Not sure how to make a group.

raytracing/__main__.py
dccote: Yes I think you had an old version

raytracing/__main__.py
dccote: I just modified it and it gets weird because -e defaults to all so we don't see the list, we just see a window popping up.  I am leaving the elif

PR #413, opened 2021-05-09
Tests examples

dccote:
* Examples directory moved into the module, but the directory was left at root level also with a note becaus the article has a direct link to the examples.
* Added a complete series of tests for all examples:  all the examples were renamed `ex\d+.py` and `fig\d\d.+.py` for the examples leading to figures in the publications.  Other files (`.png`) were copied to the directory.
* All examples now include an entry point `exampleCode()` that will get called to run the code. This is used in `testsExamples.py`.

* This will list all examples (with numbers):
```
python -m raytracing -l

All examples
==============
 1. ex01.py A single lens f = 50 mm, infinite diameter
 2. ex02.py Two lenses, infinite diameters
 3. ex03.py Finite-diameter lens
 4. ex04.py Aperture behind lens acting as Field Stop
 5. ex05.py Simple microscope system
 6. ex06.py Kohler illumination
 7. ex07.py Focussing through a dielectric slab
 8. ex08.py Virtual image at -f with object at f/2
 9. ex09.py Infinite telecentric 4f telescope
10. ex10.py Retrofocus $f_e$={0:.1f} cm, and BFL={1:.1f}
11. ex11.py Thick diverging lens computed from the Lensmaker equation
12. ex12.py Thick diverging lens built from individual elements
13. ex13.py Obtain the forward and backward conjugates
14. ex14.py Generic objectives
15. ex15.py Model Olympus objective LUMPlanFL40X
16. ex16.py Commercial doublets from Thorlabs and Edmund
17. ex17.py An optical system with vendor lenses
18. ex18.py Laser beam and vendor lenses
19. ex19.py Cavity round trip and calculated laser modes
20. fig6-lsmConfocalpinhole.py Optical section and depth of field from pinhole
21. fig7-kohler.py Kohler and critical (i.e. non-Kohler) illumination
22. fig8-widefield.py The effect of lens diameters on collection efficiency

More examples code available in: /blablabla/raytracing/examples
```

`python -m raytracing -e all` will run all examples.
`python -m raytracing -e 12,14` will run examples 12 and 14.

* We still need to get the code for the overview figure 5, which is not explicitly written in a single file.
* There seem to be several examples that are redundant, and I am not sure what to do. Also, I am unable to patch matplotlib from the command-line, I may need to rewrite these examples a bit better to allow it easily.
* Finally, I fixed the `ObjectRays()` z parameter I broke with a previous commit, but I am not sure it is done properly at this point.  The API is not clear: `ObjectRays` is not really an object.


dccote:
@JLBegin I noticed you used the `z` in ObjectRays, which I thought was not ideal.  We need to discuss this. I think it leads to confusion: the ImagingPath is not using this as a real object...

dccote:
@JLBegin I tried to fix `ObjectRays`, can you please confirm I did it right?

JLBegin (approved):

PR #425, opened 2021-07-03
GRIN lenses

dccote:
Added a new element, the GRIN lens (gradient index lens).

It is fairly trivial to add, because the matrix is known.  The difficulty was mainly because drawing the element requires to split it in smaller chunks to see the trajectory of the ray properly because it evolves as it propagates. There is now a new variable in Matrix `dz` that indicates the preferred distance with computing the transfer matrices.  They all are `0` or `self.L` except for the GRIN.

I had to modify a few tests for `testsMatrix.py` because they made assumptions on the rayTraces but really should not assume any number of rays.  Only start and end are predictable.

### DIFF
--- a/raytracing/examples/ex14.py
+++ b/raytracing/examples/ex14.py
@@ -22,7 +22,7 @@ def exampleCode(comments=None):
     path.append(obj)
     path.append(Space(10))
     
-    path.displayWithObject(diameter=20, fanAngle=0.005, comments=comments)
+    path.displayWithObject(diameter=10, fanAngle=0.005, comments=comments)
 
 if __name__ == "__main__":
     exampleCode()
--- a/raytracing/examples/ex20.py
+++ b/raytracing/examples/ex20.py
@@ -1,18 +1,20 @@
-TITLE       = "GRIN lens"
-DESCRIPTION = """
-A GRIN lens is available.  It has a quadratic index of refraction and is defined
-either with its alpha parameter or its pitch (it is easier to work with the pitch).
-Here, we have a GRIN available from Thorlabs G1P10.  Please note that the GRIN
-is not assumed to be in any medium.  Therefore, if you wish to use it in air or water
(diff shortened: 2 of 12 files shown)


dccote:
@JLBegin The drawing for the last test in `testsMatrix.py` does not render correctly, it draws the rectangle with an offset.  Can you please fix?

dccote:
Thanks. I added a fake gradient for the GRIN.

JLBegin (approved):

dccote:
@JLBegin I think I will have to trash most of what I did: I did heavy modifications in `transferMatrix` and `transferMatrices` and I think now it is more confusing and in fact, there are many errors (in tests) because I did too many things at once without good planning:
1. `transferMatrix` was modified to include a new argument: `startingAt`.  The goal was to obtain the matrix anywhere.  It felt useful.
2. For the GRIN, I really needed to have the `trace` in small steps, because the propagation direction changes everywhere. Therefore I changed the `trace` function to include a multi-step trace for the GRIN.
3. But then, I needed something general for any element of length > 0, including `Space()` to propagate in multi-step, so I added a `dz` variable to all matrices. That looks cryptic a bit.
4. I thought I could modify `transferMatrices` to return many matrices for a single element (which was the whole point of this function), but then it started creating virtual images after each propagation, so I ended up with arrows everywhere. It was not functional in practice.
5. Tons of tests started failing, because we made certain (incorrect) assumption (for instance, that there would be only 2 rays in a trace for certain elements), so changing that made tons of tests fail.
6. Finally, I figured that there was confusion about length 0 elements: if it is at z, what is the ray "at z"? Is it just before, or just after the element? I need to have some nomenclature clarification.

So my goal is to keep the `GRIN` class and the `GRINGraphic`, but trash everything else (I will keep the branch for some time, just in case). I will then start working on a real `transferMatrix/transferMatrices` refactoring.

PR #430, opened 2021-11-02
Add abs to fNumber calculation

Valeriouus:
Sometimes, got negative f/# (which is not possible). Added an abs() to fNumber in imagingPath.py.


dccote:
Duplicate

PR #432, opened 2021-11-02
Fixed negative f-number

dccote:
An f-number must be positive (ref?), therefore an absolute value was added and a test in testsIMagingPath was also added.

### DIFF
--- a/raytracing/tests/testsImagingPath.py
+++ b/raytracing/tests/testsImagingPath.py
@@ -358,8 +358,12 @@ class TestImagingPath(envtest.RaytracingTestCase):
         path = ImagingPath()
         path.append(System4f(f1=40, diameter1=25.4, f2=150, diameter2=50, label="Excitation relay"))
         path.append(Lens(diameter=16, f=500, label="ETL"))
+        
         fNumber = path.fNumber()
-        self.assertTrue(fNumber >= 0, msg="fNumber is negative")
+        focalLength, _ = path.effectiveFocalLengths()
+
+        self.assertTrue(focalLength <= 0)
+        self.assertTrue(fNumber >= 0, msg="fNumber is negative with negative lens, should be positive")
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


raytracing/tests/testsImagingPath.py
```diff
@@ -354,6 +354,25 @@ def testChiefRayInfiniteFOVNoY(self):
         with self.assertRaises(ValueError):
             path.chiefRay()
 
+    def testBugNegativeFNumber(self):
+        # Assuming:
+        path = ImagingPath()
+        path.fanAngle = 0.2094
```
dccote, line 360: Is this necessary?

dccote (commented):
Is the test as minimal as possible?
Is it clear that an f-number is never negative?

raytracing/tests/testsImagingPath.py
PyMarc2: Test passes without setting fan angle

PyMarc2: Test doesn't work with just 1 divergent lens, it returns a good test. I don't know what's the minimal system to generate a negative Number, but that combinaison of lenses seem to work.

raytracing/tests/testsImagingPath.py
```diff
@@ -354,6 +354,14 @@ def testChiefRayInfiniteFOVNoY(self):
         with self.assertRaises(ValueError):
             path.chiefRay()
 
+    def testBugNegativeFNumber(self):
+        path = ImagingPath()
+        path.append(System4f(f1=40, diameter1=25.4, f2=150, diameter2=50, label="Excitation relay"))
```
PyMarc2, line 359: quicly, this was a system that made a negative f/#. Single lenses didn't seem to work and 4f system neither.

raytracing/tests/testsImagingPath.py
dccote: But "maybe" the effectiveFocal length is not negative anymore (I know that is not the case, but we must stay rigourous).  The test should include a check that the effectiveFocalLength is indeed negative but the f-number is positive.

PyMarc2: no I tested it, It produces a negative effective focal length. Yes adding a test could be done.

PyMarc2: Oh ok I get it now ahah, yes @Valeriouus good test for you! :-)

PyMarc2: seems ready to merge now.

PR #448, opened 2022-07-24
Version checking: We timestamp even when we fail: no need to keep checking

dccote:
The module checks PyPI for a new version once per day at most.  This requires an internet connection, but if there is an error, we simply delay until the next day, no need to check again, it is not a big deal.


himbeles:
Thanks for this :)

Issue #474, opened 2024-07-29
Matplotlib update to checkBoxes


JohnKMeyer:
I was having an issue running the library with matplotlib 3.9.1, checkBoxes has no attribute rectangles or lines. these have been deprecated in version 3.7 and later.

My workaround was to either install old matplotlib or remove the lines in initVisibilityCheckBoxes that correspond to defining the boxes.

`
def initVisibilityCheckBoxes(self):
        visibility = self.visibility
        if kElementsKey in visibility.keys():
            visibility.pop(kElementsKey)

        subAxes = plt.axes([0.81, 0.4, 0.1, 0.5], frameon=False, anchor='NW')
        self.checkBoxes = CheckButtons(subAxes, visibility.keys(), visibility.values())

        # step = 0.15
        # for i, (label, rectangle, lines) in enumerate(zip(self.checkBoxes.labels,
        #                                                   self.checkBoxes.rectangles,
        #                                                   self.checkBoxes.lines)):
        #     h = 0.85 - step * i
        #     label.set_fontsize(11)
        #     rectangle.set_x(0.05)
        #     rectangle.set_y(h)
        #     rectangle.set(width=0.12, height=0.04)
        #     label.set_y(h + 0.02)
        #     label.set_x(0.2)

        #     lineA, lineB = lines
        #     lineA.set_xdata([0.05, 0.17])
        #     lineB.set_xdata([0.05, 0.17])
        #     lineA.set_ydata([h, h + 0.04])
        #     lineB.set_ydata([h + 0.04, h])

        self.checkBoxes.on_clicked(self.onCheckBoxCallback)
`

dccote:
Sorry for the delay and the kerfuffle: I merged the change without testing all examples, and the display of Achromats with their principal planes, BFl, FFL, etc.. was all screwed up, so for now I simply request matplotlib less than 3.8 (excluded) which is a terrible solution, but everything else is still fine.  We will start working on a proper fix immediately.

dccote:
Ok, it appears to have been fixed.  A new version will be released later tonight.

PR #475, opened 2024-08-02
`MplFigure.initVisibilityCheckBoxes` update for mpl3.9 compatibility

StePhanino:
Dear all, 
today I had this same issue #474.
I then tried to update the method `MplFigure.initVisibilityCheckBoxes` to complain Matplotlib3.9 `CheckButtons`.
I have a reasonable behavior (tested on examples 1,2 and 7, but I don't understand  why for example 15 something strange is happening (no checkmark for the first key).

Screenshot from 2024-08-03 01-35-18
Screenshot from 2024-08-03 01-35-29

Any idea?

### DIFF
--- a/raytracing/figure.py
+++ b/raytracing/figure.py
@@ -700,6 +700,7 @@ class MplFigure(Figure):
                 artist.set_transform(self.axes.transAxes)
             self.axes.add_artist(artist)
 
+#TODO: modify here since checkBoxes lines and rectangles attributes are deprecated
     def initVisibilityCheckBoxes(self):
         visibility = self.visibility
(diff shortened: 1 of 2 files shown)


JLBegin:
Thanks! I made a few adjustments to fix the checkbox display (their API behaves differently when setting the props during or after instantiation) and preserve naming conventions.

The only thing we are loosing is the auto-scaling "feature" of the previous box implementation. I can't seem to get it back, but it's not a big deal.

This requires to force mpl 3.7 or higher.

JLBegin (approved):

dccote (approved):

PR #480, opened 2024-08-27
Fix figure drawn twice

### LINKED ISSUE #477: At some point in time, something broke the Principal planes, EFF, BFF, FFL, display on achromat lenses
`python -m raytracing -e 16`

image

JLBegin:
Bug introduced in dd840f70ec73636576f25b82273b8f83389c43ae due to a custom `drawFigure` endpoint for the LensViewer, while the main display2D was already in charge of drawing the figure. 

I renamed the custom endpoint `createFigure` and moved the required draw call on the client side (lensViewer).

Fixes #477

### DIFF
--- a/raytracing/figure.py
+++ b/raytracing/figure.py
@@ -686,7 +686,7 @@ class MplFigure(Figure):
         for point in self.points:
             if point.hasPointMarker:
                 y = 0 if point.fixToAxis else point.y
-                self.axes.plot([point.x], [y], 'ko', markersize=3, color=point.color, linewidth=0.4)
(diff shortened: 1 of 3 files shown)


dccote (commented):
I receive a warning when running the example code 16 from `figure.py` line 689:

UserWarning [in figure.py]: color is redundantly defined by the 'color' keyword argument and the fmt string "ko" (-> color='k'). The keyword argument will take precedence.

I suggest we correct 'ko' to 'o', and also increase the marker size to 5, it looks too small.

dccote (approved):

PR #496, opened 2026-04-02
UI: turn the editor into a sequential optical layout table

dragonleopardpig:
## Summary

  This PR improves the UI editor for sequential optical layouts and fixes several issues in the `Object = Infinity` workflow.

  ## Main changes

  - make the element table editable with a constrained element-type selector
  - preserve the sequential optical layout model in the editor
  - protect special rows from invalid reordering:
    - `Object`
    - `Image`
    - the first real element when the object is at infinity
  - fix solver behavior for the image plane so the solved value is the relative image thickness
  - keep the finite image plane tied to the table state instead of silently replacing it with a solved back focal distance
  - fix plotting/runtime issues around infinity-object mode:
    - first lens remains at physical position `0`
    - object arrow is drawn at the first negative x tick
    - that tick is relabeled to `−∞`
    - incoming collimated rays originate from that object-side tick
    - image position / image thickness / red image arrow stay aligned with the current table state

  ## Result

  For infinity-object setups, the display now follows a physically consistent axis model:

  - optical elements stay on their real coordinates
  - the first lens stays at `0`
  - the object is represented on the object-side tick labeled `−∞`
  - the image plane shown in the plot and information panel matches the current editor state

  ## Notes

  This PR also removes stale display-offset behavior that caused incorrect plotting and runtime errors while editing.

  Suggested update comment on the PR:

  Updated this PR substantially on the same branch.

  The current version focuses on fixing the infinity-object workflow in the editor: the first lens stays at physical position `0`, the object is shown on the object-side tick labeled `−∞`, and the image plane handling now stays aligned with the table state.


dccote:
I am not following 100% what is happening here because the underlying machinery does not manage well infinity objects. I dont see how the UI could manage it.

The edits to the code do not follow the pedagogical spirit I am trying to maintain in the code. I ran the code and I don't understand what is happening:

image

The lens is at z=0, the image in the table appears to be at 200 which cannot be true. Changing the lens does not change this image line.  I like the idea of putting the image in the table, but this does not work at all. Not sure how the reordering would work because nothing happens when I click anywhere.

There is no unit test either for anything.


I recommend you keep working on it on your side and create a Pull Request when it is ready.

